PR #3, opened 2024-01-30
Balance Druid sim updates

kayla-glick:


sim/druid/runes.go
```diff
@@ -69,7 +69,7 @@ func (druid *Druid) applySunfire() {
 			NumberOfTicks: ticks,
 			TickLength:    time.Second * 3,
 			OnSnapshot: func(sim *core.Simulation, target *core.Unit, dot *core.Dot, _ bool) {
-				dot.SnapshotBaseDamage = (baseDotDamage / float64(ticks)) + 0.13*dot.Spell.SpellPower()
+				dot.SnapshotBaseDamage = baseDotDamage + 0.13*dot.Spell.SpellPower()
```
kayla-glick, line 72: The base damage is actually `baseCalc * .65 * 4` but up above we only use `baseCalc * .65`. This means each tick was actually dealing 1/4 the correct damage. I opted to remove the division by # of ticks rather than multiply the base dot damage and then divide here

image

sim/druid/moonfire.go
```diff
@@ -40,10 +41,10 @@ func (druid *Druid) getMoonfireBaseConfig(rank int) core.SpellConfig {
 				Label:    "Moonfire",
 				ActionID: core.ActionID{SpellID: spellId},
 			},
-			NumberOfTicks: core.TernaryInt32(rank < 2, 3, 4),
+			NumberOfTicks: ticks,
 			TickLength:    time.Second * 3,
 			OnSnapshot: func(sim *core.Simulation, target *core.Unit, dot *core.Dot, _ bool) {
-				dot.SnapshotBaseDamage = (baseDotDamage / 3.0) + spellDotCoeff*dot.Spell.SpellPower()
+				dot.SnapshotBaseDamage = (baseDotDamage / float64(ticks)) + spellDotCoeff*dot.Spell.SpellPower()
```
kayla-glick, line 47: This should've been dividing by the number of ticks

sim/core/database_load.go
```diff
@@ -42,5 +43,13 @@ func init() {
 		}
 	}
 
+	for i, suffix := range db.RandomSuffixes {
+		simDB.RandomSuffixes[i] = &proto.ItemRandomSuffix{
+			Id:    suffix.Id,
+			Name:  suffix.Name,
+			Stats: suffix.Stats,
+		}
+	}
```
kayla-glick, line 52: Suffixes needed to be loaded into the DB otherwise it would crash when using a suffix item in a test.

ui/core/components/inputs/buffs_debuffs.ts
```diff
@@ -314,12 +314,14 @@ export const RAID_BUFFS_CONFIG = [
...
-		stats: []
+		stats: [],
+		faction: Faction.Alliance,
 	},
 	{
 		config: AllStatsPercentBuffHorde,
 		picker: IconPicker,
-		stats: []
+		stats: [],
+		faction: Faction.Horde,
```
kayla-glick, line 324: Aspect of the lion being shared by two active pickers where one was hidden was causing the value to get forcibly disabled. This will filter one out all the time to prevent the issue

PR #73, opened 2024-02-06
Stormstrike only hits with OH when Dual Wield Spec rune is active

Foknetics:


kayla-glick (approved):
Totally forgot this was a thing. This is why I'm an Ele main 😅

PR #98, opened 2024-02-07
Add Electromantic Devastator's Mail set bonus

Foknetics:
Looking for any help on verifying the `OnSpellHitDealt` aspect is correct. I modeled it after the Stormshroud proc but that one notably has a `|| !spell.ProcMask.Matches(core.ProcMaskMelee)` to fail out of the proc. The differences being Stormshroud's tooltip reads "on melee attack" where as this set reads as "on attack". Is there somewhere we can verify the interaction (item db?) before someone has the set?

### DIFF
--- a/sim/common/sod/melee_sets.go
+++ b/sim/common/sod/melee_sets.go
@@ -130,6 +130,7 @@ var ItemSetElectromanticDevastator = core.NewItemSet(core.ItemSet{
 					c.AddMana(sim, 100, metrics)
 				},
 			})
+			procChance := 0.05
 			c.RegisterAura(core.Aura{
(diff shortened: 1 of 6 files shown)


TheGroxEmpire:
The proc for this aura appears to be similar to Judgement of Wisdom. You can refer to the wotlk implementation of it. It should also proc on miss and there are some other special cases listed here:

https://github.com/wowsims/wotlk/blob/master/sim/core/debuffs.go#L202

Foknetics:
Changes should be good to go now modelled after the JoW reference

TheGroxEmpire (approved):
LGTM!

PR #168, opened 2024-02-12
Adjust some level 40 agi/int crit mults

FelixPflaum:
Added level 40 agi and int crit mults for Warrior, Rogue, Druid, Hunter and Shaman. Druid is first-hand data, others are what my WA sent me from guild members, but should be correct too.

Also removed the global int crit scaling from mana.go (was remnant WotLK code). Moved it to warlock/pet.go to no alter behavior there, as that was the only place that probably still needs that stat dependency. Added a TODO comment there because in any case the value will not be correct.


FelixPflaum:
Guess I need to update the tests for new values too?

rosenrusinov:
> Guess I need to update the tests for new values too?

You need to merge latest master in your branch and then update tests due to recent changes

PR #170, opened 2024-02-12
Fix attack table miss, add hit supression, fix crit supression

FelixPflaum:
Attack table had the wrong if condition for miss chance and always used the same case as a result. Additionally I fixed hit suppression to only apply on actual gained hit chance and not behave as additional miss chance. This only affects level +3 cases.
See https://github.com/magey/classic-warrior/wiki/Attack-table#miss as reference.

Crit suppression only had the case for target being the same level or lower implemented. Added case for target being higher level and use that conditionally.
See https://github.com/magey/classic-warrior/wiki/Attack-table#critical-strike

A quick test (auto attack only feral, 12.69% crit) results in 
| Value/Level diff | 0 | 1 | 2 | 3 |
| - | - | - | - | - |
| Crit % | 12.67 | 11.7 | 10.6 | 7.83 |
| Miss % | 4.98 | 5.53 | 6.02 | 8.06 |
| Miss % with 1% hit from gear | 3.96 | 4.49 | 5.04 | 8.02 |

which is expected.

The 1.8% aura crit suppression for targets 3+ levels above the player is currently always applied, even if the player does not have any crit from auras. Fixing this would require adding a new stat to separate "base crit" from "aura crit". Not having at least 2% crit from talents, gear or other auras should be an edge case once bosses are +3 level, but it's still not technically correct to simply add 0.018 to MeleeCritSuppression. Thoughts on that?

Because of that and because test results need an update anyways I leave it as a draft for now.


kayla-glick:
Hey @FelixPflaum would you mind pulling down the latest and updating the tests? This will be super helpful for making the sims more accurate

FelixPflaum:
Sure, was just still unsure about the aura crit supression. I added a comment there in case anyone is ever bothered enough to change that :D

sim/core/target.go
```diff
@@ -219,24 +220,39 @@ func NewAttackTable(attacker *Unit, defender *Unit) *AttackTable {
...
+		// TODO (maybe): This is technically not correct, but it shouldn't matter outside of edge cases.
+		// These 1.8% should only be subtracted from crit chance gained from auras,
+		// i.e. talents, gear and buffs, NOT base crit and crit from agility!
+		// See https://github.com/magey/classic-warrior/wiki/Attack-table#critical-strike
+		// That means if a character with <2% crit from auras attacks a +3 level target the sim will be wrong.
+		// The chance of that being the case once bosses are +3 in SoD should be very small though.
+		// Most (all?) affected specs have crit in their talents to begin with.
 		if (defender.Level - attacker.Level) >= 3 {
 			table.MeleeCritSuppression += 0.018
 		}
```
kayla-glick, line 258: Hey @rosenrusinov if you get a chance would you mind taking a look at this? I'm not familiar at all with crit suppression mechanics

FelixPflaum:
I asked about it in discord too. Can't remember who it was right now, but the reply I got was basically that it's probably not worth it. Maybe I'm missing something and there is an easy obvious way to know where crit comes from, but because the sim doesn't actually emulate the game's aura system and everything is just added to the MeleeCrit stat I don't think there is.

sim/core/target.go
rosenrusinov: Its fine to be merged as it is for now. We are 2 phases away from this having any significance so we have time to figure out how to deal with it when its relevant

kayla-glick:
Dang it I'm so sorry @FelixPflaum can you update tests one more time? I'll merge this after that

kayla-glick (approved):

PR #171, opened 2024-02-12
Feral P2

FelixPflaum:
Implement P2 spells and changes.

- [x] Level 40 claw base damage
- [x] Mangle and Shred weapon mult change
- [x] Rake with AP scaling [^1]
- [x] Berserk rune
- [x] King of the Jungle rune
- [x] Ferocious Bite
- [x] Innervate
- [x] Automatic Crowd Pummeler
- [x] Iron Counterweight
- [x] Blood Frenzy talent

[^1]: From my testing AP coef seems to be 4% on both direct damage and per tick.


sim/druid/ferocious_bite.go
```diff
@@ -0,0 +1,125 @@
...
+	// Implement this spell as having multiple ranks available later? Like e.g. Starfire?
+	// Or just uncomment this once the book is available?
+	// {
+	// 	id:           31018,
+	// 	level:        60,
+	// 	dmgBase:      52.0,
+	// 	dmgRange:     60.0,
+	// 	dmgPerCombo:  147.0,
+	// 	dmgPerEnergy: 2.7,
+	// },
```
kayla-glick, line 61: I think it should be ok to un-comment? The spells only get registered if the level is below the character's level

kayla-glick: Also PS I really like the formatting you used here

sim/druid/rake.go
```diff
@@ -0,0 +1,138 @@
...
+
+func (druid *Druid) registerRakeSpell() {
+	// Add highest available rake rank for level.
+	for rank := len(rakeSpells) - 1; rank >= 0; rank-- {
+		if druid.Level >= rakeSpells[rank].level {
+			config := druid.newRakeSpellConfig(rakeSpells[rank])
+			druid.Rake = druid.RegisterSpell(Cat, config)
+			return
+		}
+	}
```
kayla-glick, line 60: Instead of iterating through the ranks like this, you could get the correct rank information from the start using a map of levels to ranks like this:
```
rankConfig := map[int32]int32{
  25: 1,
  40: 2,
  50: 3,
  60: 4,
}[druid.Level]
```

sim/druid/ferocious_bite.go
FelixPflaum: No clue what they will do with SoD, but it's likely not available immediately on 60. Those books were made available with AQ, or DM, can't remember right now.

sim/druid/rake.go
FelixPflaum: I intentionally tried to avoid hardcoding levels so it's theoretically working with arbitrary levels, not just 25, 40, 50 and 60. I know that realistically those levels are the only ones ever used, but it still doesn't feel right to me to restrict it for no reason. Should I change it?

kayla-glick: Fair enough!

kayla-glick (approved):
:shipit:

FelixPflaum:
@kayla-glick Noticed that Rip values were off after you changed the tick count. I changed Rip definitions to be dmg per tick so this can't happen again. Did the same for Rake because why not. Also added missing aura labels for Rip and Rake. Otherwise it's still the same since you reviewed it.

PR #208, opened 2024-02-16
Shaman 2/20 changes

Foknetics:
https://us.forums.blizzard.com/en/wow/t/season-of-discovery-class-tuning-incoming-february-20/1788506

### DIFF
--- a/sim/shaman/runes.go
+++ b/sim/shaman/runes.go
@@ -44,6 +44,8 @@ func (shaman *Shaman) applyDualWieldSpec() {
 		return
 	}
 
+	shaman.AutoAttacks.OHConfig().DamageMultiplier *= 1.5
+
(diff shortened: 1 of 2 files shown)


sim/shaman/runes.go
```diff
@@ -146,7 +148,7 @@ func (shaman *Shaman) applyMaelstromWeapon() {
 	buffSpellId := 408505
 	buffDuration := time.Second * 30
 
-	ppm := 10.0
+	ppm := 15.0
```
kayla-glick, line 151: This is only true if the main hand is imbued with Windufury

Foknetics: Good call out, forgot that piece. Implemented it similar to our Lava Lash Flametongue code.

kayla-glick (approved):
:shipit: ✨

PR #214, opened 2024-02-16
Weapon Skill Refactor

TheGroxEmpire:
- Weapon Skills are now correctly implemented for each weapon slot
- Attack Tables now are 2 dimensional (``AttackTables[UnitIndex][CastType]``). First dimension is ``UnitIndex``, second dimension is the ``CastType``.
- ``CastType`` is the slot in which spell is cast from. There are three types of ``CastType``:  ``CastTypeMainHand``, ``CastTypeOffHand``, ``CastTypeRanged``
- Spells are in default cast from ``CastTypeMainHand``
- Spells ``CastType`` can be adjusted with ``SpellConfig``
- Added Wowhead tooltip parsing for weapon skills

Note to other Class Developers:
- Please make sure to change your spell ``CastType`` if needed. Most Classes shouldn't be affected as most spells are from ``MainHand`` (this is the default Attack Table as before). However, class that has spells that are being cast from ranged or offhand such as Hunter should update those spells ``CastType`` accordingly.


kayla-glick (approved):

kayla-glick (approved):
Tentative approval because attack tables still go above my head 😅

PR #302, opened 2024-02-25
Add ret sim

hypernormalisation:
Places where the core code/code for other specs was touched:

- an APLAction and an APLValue to support a "primary seal" were added, to allow the user to easily swap seals on the fly without multiple changes to their APL.
- the Sanctity Aura buff was implemented in the sim and UI, and added to the test suite for p2 buffs (this buff only appears on the paladin and priest sim frontends).
- in accordance with the above, fixed a typo in the shadow priest test suite that was incorrectly loading p1 gear/rotation for p2 testing, and therefore producing inconsistent test results with the Sanc aura inclusion. These tests have been redone, all other spec tests pass.
- some holy paladin and prot paladin code was altered to support the new paladin protos.
- retribution sim was registered and set at alpha launch status.

Implementation of the ret sim includes almost all relevant talents and abilities. The notable missing debuff is Judgement of the Crusader, which per a discussion in the discord might be best supported by making alterations to the core sim code. A "Known Issue" is included in the frontend to suggest a workaround for users to estimate the dps value of the debuff in the meantime by manipulating their spellpower stat.

Gear/talent presets are provided for p1 and p2, including some support for the "shockadin" dps spec. Unified APLs for p1 and p2 that are viable across traditional ret/shockadin are supplied.


sim/paladin/apl_values.go
```diff
@@ -0,0 +1,93 @@
...
+	paladin := action.paladin
+	// If the player options are incorrectly configured, then no primary seal will be selected.
+	if paladin.PrimarySealSpell == nil {
+		return
+	}
+	paladin.PrimarySealSpell.Cast(sim, paladin.CurrentTarget)
+}
+
+func (action *APLActionCastPaladinPrimarySeal) IsReady(sim *core.Simulation) bool {
+	return sim.CurrentTime > action.lastAction
```
rosenrusinov, line 84: You probably want to add `&& paladin.PrimarySealSpell.CanCast(sim)` so this action is not executed when the spell cant be cast

sim/paladin/crusader_strike.go
```diff
@@ -0,0 +1,45 @@
...
+				Timer:    paladin.NewTimer(),
+				Duration: time.Second * 6,
+			},
+		},
+		DamageMultiplier: 0.75,
+		CritMultiplier:   paladin.MeleeCritMultiplier(),
+		ThreatMultiplier: 1,
+
+		ApplyEffects: func(sim *core.Simulation, target *core.Unit, spell *core.Spell) {
+			baseDamage := spell.Unit.MHNormalizedWeaponDamage(sim, spell.MeleeAttackPower()) + spell.BonusWeaponDamage()
```
rosenrusinov, line 40: `spell.BonusWeaponDamage()` is usually applied after any coefficients to weapon attacks so you probably want to move the 0.75 multi to here. (I could be wrong dont know if this spell is not an exception)

sim/paladin/divine_favor.go
```diff
@@ -0,0 +1,69 @@
...
+				spell.BonusCritRating -= core.CritRatingPerCritChance * 100
+			})
+		},
+		OnSpellHitDealt: func(aura *core.Aura, sim *core.Simulation, spell *core.Spell, result *core.SpellResult) {
+			if spell.SpellCode != SpellCode_PaladinHolyShock {
+				return
+			}
+			// Remove the buff and put skill on CD
+			aura.Deactivate(sim)
+			cdTimer.Set(sim.CurrentTime + cd)
```
rosenrusinov, line 47: you should not need to set this timer manually if you're already passing it as the CD timer of the spell

sim/paladin/divine_storm.go
```diff
@@ -0,0 +1,66 @@
...
+		},
+
+		DamageMultiplier: 1.1,
+		CritMultiplier:   paladin.MeleeCritMultiplier(),
+		ThreatMultiplier: 1,
+		ApplyEffects: func(sim *core.Simulation, target *core.Unit, spell *core.Spell) {
+			curTarget := target
+			totalDamageDealt := 0.0
+			for hitIndex := int32(0); hitIndex < numHits; hitIndex++ {
+				baseDamage := spell.Unit.MHWeaponDamage(sim, spell.MeleeAttackPower()) + spell.BonusWeaponDamage()
```
rosenrusinov, line 52: same concern as CS with the weapon strike mode being applied to bonus weapon damage

sim/paladin/exorcism.go
```diff
@@ -0,0 +1,112 @@
...
+		CritMultiplier:   paladin.SpellCritMultiplier(),
+		BonusCritRating:  paladin.getBonusCritChanceFromHolyPower(),
+
+		ApplyEffects: func(sim *core.Simulation, target *core.Unit, spell *core.Spell) {
+			baseDamage := sim.Roll(baseDamageMin, baseDamageMax) + spellCoeff*spell.SpellDamage()
+
+			bonusCrit := core.TernaryFloat64(
+				guaranteed_crit,
+				100*core.CritRatingPerCritChance,
+				0)
```
rosenrusinov, line 71: This calculation can be moved up and out of the apply effects func

rosenrusinov (commented):
i did a quick glance and left some comments i'll do a proper look tomorrow

sim/paladin/divine_favor.go
hypernormalisation: The Divine Favor spell only goes on cooldown once the Divine Favor aura expires, i.e. the player spends it on a Holy Shock cast.

The point of this code is to only set the cooldown running on the Divine Favor spell when the Holy Shock is cast. If I remove this line of code, Divine Favor gets cast every 2 mins on cooldown regardless of when the aura actually expired.

Is there a more elegant way to attain this behaviour that I have missed?

rosenrusinov: No in this case what you did makes sense

sim/paladin/paladin.go
```diff
@@ -0,0 +1,175 @@
...
+	paladin.AddStatDependency(stats.Intellect, stats.SpellCrit, core.CritPerIntAtLevel[character.Class][int(paladin.Level)]*core.SpellCritRatingPerCritChance)
+
+	// Paladins get 0.0167 dodge per agi. ~1% per 59.88
+	paladin.AddStatDependency(stats.Agility, stats.Dodge, (1.0/59.88)*core.DodgeRatingPerDodgeChance)
+
+	// Paladins get 1 block value per 20 str
+	paladin.AddStatDependency(stats.Strength, stats.BlockValue, .05)
+
+	// Bonus Armor and Armor are treated identically for Paladins
+	paladin.AddStatDependency(stats.BonusArmor, stats.Armor, 1)
```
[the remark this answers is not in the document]
hypernormalisation: Not intentional at all, this was copied (early on) from the live wotlk code:
https://github.com/wowsims/wotlk/blob/master/sim/paladin/paladin.go#L194

I had assumed this was how to get bonus armor accounted for properly. It's this way in the druid and shaman implementations for SoD too, e.g:
https://github.com/wowsims/sod/blob/master/sim/shaman/shaman.go#L36

Should this line be removed?

sim/paladin/paladin.go
```diff
@@ -0,0 +1,175 @@
...
+	core.FillTalentsProto(paladin.Talents.ProtoReflect(), talentsStr, TalentTreeSizes)
+
+	paladin.PseudoStats.CanParry = true
+	paladin.EnableManaBar()
+	paladin.AddStatDependency(stats.Strength, stats.AttackPower, 2.0)
+	paladin.AddStatDependency(stats.Agility, stats.MeleeCrit, core.CritPerAgiAtLevel[character.Class][int(paladin.Level)]*core.CritRatingPerCritChance)
+	paladin.AddStatDependency(stats.Intellect, stats.SpellCrit, core.CritPerIntAtLevel[character.Class][int(paladin.Level)]*core.SpellCritRatingPerCritChance)
+
+	// Paladins get 0.0167 dodge per agi. ~1% per 59.88
+	paladin.AddStatDependency(stats.Agility, stats.Dodge, (1.0/59.88)*core.DodgeRatingPerDodgeChance)
```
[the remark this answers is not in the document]
hypernormalisation: I'm informed that it does scale with level in classic. 

e just generally seem to be missing a proper `CritPerAgiAtLevel` table analogue for dodge in the SoD code at present, no other specs, not even those with tank specs enabled, implement this at present.

Also, if this data is trustworthy it would imply the BaseDodge is wrong a few lines later:
https://www.wowhead.com/classic/guide/classic-wow-stats-and-attributes-overview#dodge

I can get someone to test this in-game.

hypernormalisation:
Have spotted an error affecting the 2h weapon spec talent, that applies a % damage modifier, so that shouldn't be applying a global damage done bonus, it should be affecting physical damage and anything with melee defense type.

Will sort that before we proceed.

sim/paladin/paladin.go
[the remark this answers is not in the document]
rosenrusinov: You seem to be right and for some reason the BonusArmor stat is not included in any base calculations and the above call makes it being accounted for. This seems a bit weird to me to not be just part of the Mitigation function for Armor instead but I wont be changing anything for now so I guess what you're doing here should stay

sim/paladin/paladin.go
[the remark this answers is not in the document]
rosenrusinov: I think this link is only for level 60. You can find stats from the RatingBuster addong that VoraciousGhost(in the sim discord) maintains. We have used his data for the crit ratings formulas

sim/paladin/paladin.go
```diff
@@ -0,0 +1,175 @@
...
+func (paladin *Paladin) AddRaidBuffs(raidBuffs *proto.RaidBuffs) {
+	// Buffs are handled explicitly through APLs now
+}
+
+func (paladin *Paladin) AddPartyBuffs(_ *proto.PartyBuffs) {
+}
+
+func (paladin *Paladin) Initialize() {
+	// Update auto crit multipliers now that we have the targets.
+	paladin.AutoAttacks.MHConfig().CritMultiplier = paladin.MeleeCritMultiplier()
```
rosenrusinov, line 91: I think this might be just a copy/paste leftover as I see you have no requirements to target in your CritMultiplier

sim/paladin/retribution/retribution.go
```diff
@@ -0,0 +1,79 @@
...
+
+	ret := &RetributionPaladin{
+		Paladin:     pal,
+		PrimarySeal: retOptions.Options.PrimarySeal,
+	}
+
+	// ret.PaladinAura = retOptions.Options.Aura
+
+	ret.EnableAutoAttacks(ret, core.AutoAttackOptions{
+		MainHand:       ret.WeaponFromMainHand(0), // Set crit multiplier later when we have targets.
```
rosenrusinov, line 39: You can set Crit Multiplier alraedy as it does not depend on target

sim/paladin/som.go
```diff
@@ -0,0 +1,101 @@
...
+			},
+		},
+
+		ApplyEffects: func(sim *core.Simulation, _ *core.Unit, spell *core.Spell) {
+			if paladin.CurrentSeal != nil {
+				paladin.CurrentSeal.Deactivate(sim)
+			}
+			paladin.CurrentSeal = aura
+			paladin.CurrentSeal.Activate(sim)
+			paladin.CurrentSealExpiration = sim.CurrentTime + SealDuration
```
rosenrusinov, line 98: This looks like something you do in all seal spells. Can you make it a generic function in Paladin something like:
```
func (paladin *Paladin) ApplySeal(aura *core.Aura) {
	if paladin.CurrentSeal != nil {
		paladin.CurrentSeal.Deactivate(sim)
	}
	paladin.CurrentSeal = aura
	paladin.CurrentSeal.Activate(sim)
	paladin.CurrentSealExpiration = sim.CurrentTime + aura.Duration
}
```
And then call this from all seal ApplyEffects

ui/core/components/inputs/buffs_debuffs.ts
```diff
@@ -398,6 +398,16 @@ export const RetributionAura = withLabel(
 	}),
 	'Retribution Aura',
 );
+export const SanctityAura= withLabel(
+	makeBooleanRaidBuffInput({
+		actionId: (player) => player.getMatchingSpellActionId([
+			{ id: 20218, minLevel: 30 },
+		]),
+		fieldName: 'sanctityAura',
+		showWhen: (player) => ((player.getFaction() == Faction.Alliance) && ((player.getClass() == Class.ClassPaladin) || (player.getClass() == Class.ClassPriest))),
```
rosenrusinov, line 407: leave only the faction check in here. The class checks should not be here but instead when adding this buff to the main buff/debuffs config it should be configured to be based on HolyPower stats (instead of empty stat array as you have it). This will propery add this buff to every sim that has HolyPower in its EPs

ui/core/components/inputs/buffs_debuffs.ts
```diff
@@ -937,6 +947,12 @@ export const RAID_BUFFS_CONFIG = [
 		stats: [Stat.StatArmor],
 		faction: Faction.Alliance,
 	},
+	{
+		config: SanctityAura,
+		picker: IconPicker,
+		stats: [],
```
rosenrusinov, line 953: Stat.HolyPower

sim/paladin/retribution/retribution_test.go
```diff
@@ -0,0 +1,226 @@
+package retribution
+
+import (
+	"testing"
+
+	_ "github.com/wowsims/sod/sim/common" // imported to get item effects included.
```
rosenrusinov, line 6: Item effects have their own tests so I dont feel this is needed here

sim/paladin/soc.go
```diff
@@ -0,0 +1,157 @@
...
+		ProcMask:         core.ProcMaskMeleeMHSpecial | core.ProcMaskSuppressedExtraAttackAura, // This needs figured out properly
+		Flags:            core.SpellFlagMeleeMetrics,
+		RequiredLevel:    level,
+		DamageMultiplier: 0.7,
+		ThreatMultiplier: 1.0,
+		CritMultiplier:   paladin.MeleeCritMultiplier(),
+
+		ApplyEffects: func(sim *core.Simulation, target *core.Unit, spell *core.Spell) {
+			baseDamage := -1 + spell.Unit.MHWeaponDamage(sim, spell.MeleeAttackPower())
+			fullDamage := baseDamage + socProcSpellCoeff*spell.SpellPower() + spell.BonusWeaponDamage()
```
rosenrusinov, line 76: SpellPower instead of SpellDamage seems to be wrong

ui/core/components/inputs/buffs_debuffs.ts
rosenrusinov: Actually you dont even need the faction check as well as it looks like something that is also handled from the other configuration

sim/paladin/crusader_strike.go
rosenrusinov: After checking with @TheGroxEmpire this seems correct

sim/paladin/soc.go
hypernormalisation: Fixed in https://github.com/wowsims/sod/pull/302/commits/553a78db7093e4aeba162e7b6a5ed6686bc1cf68

sim/paladin/retribution/retribution_test.go
hypernormalisation: Fixed in https://github.com/wowsims/sod/pull/302/commits/4b628b513d4a78739a4105be686f8a750f920c29

ui/core/components/inputs/buffs_debuffs.ts
hypernormalisation: Fixed in https://github.com/wowsims/sod/pull/302/commits/6085426bc9c220f0b23d0f0779006ae7408a9b60

sim/paladin/paladin.go
hypernormalisation: Correct. So this should just be
`	paladin.AutoAttacks.MHConfig().CritMultiplier = paladin.DefaultMeleeCritMultiplier()` 
and cutting out the defunct function call, right?

sim/paladin/retribution/retribution.go
hypernormalisation: Ah, I see this interacts with the above feedback. Given the crit multiplier on melee swings is constant between prot and ret, should it be set in paladin.go like we describe above, or in the individual retribution.go and eventually protection.go files?

sim/paladin/som.go
hypernormalisation: Implemented in https://github.com/wowsims/sod/pull/302/commits/0663be7b6d346fbe5b525568cde39b19ddbb44db

sim/paladin/paladin.go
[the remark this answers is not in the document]
hypernormalisation: https://github.com/raethkcj/RatingBuster/blob/master/libs/StatLogic/Vanilla_Logic.lua#L121

Relevant line, appears to be the same as `CritPerAgiAtLevel` for paladins. It seems like this should have its own function to handle special cases for rogues and be human readable.

I could change this to
`	paladin.AddStatDependency(stats.Agility, stats.Dodge, core.CritPerAgiAtLevel[character.Class][int(paladin.Level)]*core.DodgeRatingPerDodgeChance)`

or, given we're not implementing the prot sim atm where avoidance stats are actually relevant, we could just comment out the stuff related to basedodge/baseparry and the dodge agi dependency with relevant comments; then we'd add this when there's an appropriately named function in core like `DodgePerAgiAtLevel` to prevent confusion.

What do you think?

sim/paladin/paladin.go
rosenrusinov: yes

sim/paladin/retribution/retribution.go
rosenrusinov: I think we keep auto attack initialization in the specific specs

sim/paladin/paladin.go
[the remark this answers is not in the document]
rosenrusinov: you can comment it out for now if its not relevant yes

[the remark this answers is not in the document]
hypernormalisation: Addressed in https://github.com/wowsims/sod/pull/302/commits/0cd88fb8da4ba855a85b24ec39f2aab83d28f550

sim/paladin/retribution/retribution.go
hypernormalisation: handled in https://github.com/wowsims/sod/pull/302/commits/eeaee60237a3194c07a710dc6ac986c583373ef8

hypernormalisation:
> Have spotted an error affecting the 2h weapon spec talent, that applies a % physical damage modifier, so that shouldn't be applying a global damage done bonus, it should be affecting physical damage and anything with melee defense type that doesn't have the no active defense attribute _only_.
> 
> Pride also clarified some things about Judgement spells in classic, I'd incorrectly assumed all went through a dummy spell but that's just Judgement of Command. Will need to modify the base judgement, JoC, and JoR and how they roll to hit/miss.
> 
> Will sort those before we proceed.

These should now have been dealt with in:
- https://github.com/wowsims/sod/pull/302/commits/1943718b3a44493cd93fd87a8b907a9ce973d5b7
- https://github.com/wowsims/sod/pull/302/commits/7e37610b4124bda0e09d42f2be431460b0306084
- https://github.com/wowsims/sod/pull/302/commits/5aa1c13286de9c5b3b6bd407f91fe417ac70fb82

sim/paladin/soc.go
```diff
@@ -0,0 +1,168 @@
...
+	level := socLevels[rank]
+	scalingLevelMin := socLevelMinMaxEffects[rank][0]
+	scalingLevelMax := socLevelMinMaxEffects[rank][1]
+	judgeBasePoints := socJudgeBasePoints[rank]
+	judgePointsPerLevel := socJudgeRealPointsPerLevel[rank]
+	judgeDieSides := socEffectDieSides[rank]
+
+	levelsToScale := min(paladin.Level, scalingLevelMax) - scalingLevelMin
+	judgeMinDamage := judgeBasePoints + float64(levelsToScale)*judgePointsPerLevel
+	judgeMaxDamage := judgeMinDamage + judgeDieSides
```
rosenrusinov, line 57: I notice this in a couple of places now and it seems to be standard for some abilities. Do you think its better to move it into a utility function somewhere in core (maybe spell.go)

hypernormalisation: I agree with that.

When figuring out the damage/damage range on a spell at any particular level, directly inputting the SpellEffect table data and running the relevant calc in the code is imo favourable to the sim implementer manually calculating a min/max at the relevent level brackets.

I've seen at least one instance in the code of a spell's base damage not being properly implemented for the level, having the formula be explicit like this helps in external validation.

PR #314, opened 2024-02-27
Revamp Results Log UI

kayla-glick:
image

image


rosenrusinov:
Damage events and Major Cooldowns could also be linked to the spell id
image
image

rosenrusinov:
Also do you think it would be nicer to make all Number values bold? (Like damage/threat/stacks/etc.)

kayla-glick:
Nice catch thanks @rosenrusinov!

kayla-glick:
@rosenrusinov I did you one better and added colors for resources and damage
image
image

kayla-glick:
I also noticed in that last screenshot that aura stacks needed to be updated to use an Action ID link

rosenrusinov:
This is such a huge improvement from how this screen was. Amazing 🤩

PR #342, opened 2024-03-04
Multi school support

FelixPflaum:
Add support for multi school spells.
Also fixes resistance (level based resistance was incorrect, cap was based on defender instead of attacker) and adds some tests for that.

## Multi school mechanics this is based on:
- Choose lowest resistance (armor seems to be chosen when it is somewhat below resistance, exact mechanic unknown atm, see [this sheet](https://docs.google.com/spreadsheets/d/e/2PACX-1vQirvUMdrJZUaafUeLnO_Vwu-WXwBTi19EtO0xJkAIo_Y-Kn3a7Im1HzSLDOVquYEzEYFhI-lk9qE3Q/pubhtml))
- Choose highest spell power stat.
- Choose highest school damage dealt modifier on caster.
- Choose highest school damage taken modifier on target.

The latter 2 points can be checked in logs. I looked at
- Living Flame: Does only benefit from one of the effects of Mekkatorque's Arcano-Shredder
- Mind Spike: For this there can be 2 distinct spells, Shadow Vulnerability and CoE. Considering the dmg range I can't say with 100% certainty that it does not benefit from both, but it really seems like it doesn't.
- Bloodbark Crusher (Druid STV mace): The on use does 5 physical + nature damage. Using SR/TF (phys only) and PI (magic only) it becomes obvious that it does not stack, as seen in e.g. this log: https://vanilla.warcraftlogs.com/reports/jFxKvyb3tgc14GVY#boss=-3&difficulty=0&type=damage-done&source=1&ability=436481&view=events (First 2 trash fights I have PI)
- Same mace same log: Hits for 10-11 due to SR, TF and DMF stacking phys on me, Dreamstate/SS providing 20% nature on target.

## Implementation

I created additional pseudo school indices for multi-schools. Spells are defined with their full school mask, which is translated to the respective (pseudo) school index on spell creation. LUTs are created on startup to then lookup composing indices for multi schools at runtime. Everything is then still an array lookup wherever possible. Full school iterations can be prevented.

### Resistance
Considering the amount of ms spells and the work that would be needed to change the way resistance stats are handled, I just reworked the resistance switch to use one of the mentioned LUTs to find the lowest resistance in the case of a ms school index. The performance impact is practically non-existent anyway.

A short test ingame hints towards armor only being chosen if close or lower than the next lowest resistance choice available. The current implementation simply uses armor if it's the lowest. Will require further testing if relevant spells get added to the game.

### Spell power
Same as resistance, just instead finds the highest. Same considerations apply here. Due to the way stats are implemented the spell power selection is kinda ugly, but it works.

**TODO for later**: Bonus damage taken, i.e. extra spell power from target debuffs, is currently only implemented for physical. I'll probably change that after this PR, would bloat his quite a lot.

### School modifiers
While it would be somewhat cleaner to update multipliers on change, it would most likely add more overhead than it can save as long as ms spells are a small minority. Choosing the best modifiers on spell usage seems to be the way to go for now. This happens in spell_result.go code depending on the spell's school.

Note: From very rough counting there are 10-20x as many reads on the school mods as there are writes to them. Multi schools would need to make up a considerable amount of dmg events to make the computation shift to on-write even potentially worth it.

## Benchmark results
Benchmark comparing this to master as of commit f2374dced9a57326dc1a7d1aa1c1bab6eca8124c: <https://pastebin.com/raw/n3AMK3sv>
(Using script and changes from my benchmark branch: <https://github.com/FelixPflaum/sod/tree/benchmarks>)


sim/mage/talents.go
```diff
@@ -23,6 +23,7 @@ func (mage *Mage) ApplyTalents() {
 	if mage.Talents.ElementalPrecision > 0 {
 		bonusHit := 2 * float64(mage.Talents.ElementalPrecision) * core.SpellHitRatingPerHitChance
 		mage.OnSpellRegistered(func(spell *core.Spell) {
+			// TODO: This is not correct, it will affect non-class spells this way.
```
kayla-glick, line 26: Just FYI all of these are updated in the `mage` branch so these TODOs aren't needed 😄

FelixPflaum: Removed them again.

sim/core/spell_result.go
```diff
@@ -457,6 +488,7 @@ func (result *SpellResult) applyTargetModifiers(spell *Spell, attackTable *Attac
 		return
 	}
 
+	// TODO: Add other schools. Multischools should then chose highest.
```
kayla-glick, line 491: Is this still TODO?

sim/core/spell_school.go
```diff
@@ -0,0 +1,256 @@
...
+	case proto.SpellSchool_SpellSchoolFrost:
+		return SpellSchoolFrost
+	case proto.SpellSchool_SpellSchoolHoly:
+		return SpellSchoolHoly
+	case proto.SpellSchool_SpellSchoolNature:
+		return SpellSchoolNature
+	case proto.SpellSchool_SpellSchoolShadow:
+		return SpellSchoolShadow
+	default:
+		return SpellSchoolPhysical
```
kayla-glick, line 187: 🔨 This feels like one of those cases where physical being the zero value in the proto enum could cause unexpected problems at some point. What do you think about adding a `SpellSchoolUnknown` value as the first entry in that enum?

kayla-glick (commented):
Just a ❓ or 2 but otherwise this is AWESOME! Really appreciate your work on it. It looks like DPS results went down very slightly across the board. Any ideas what might've caused that?

sim/core/spell_result.go
FelixPflaum: Yes. Implementing it most likely requires adding the coefficient to all spells, because it would be required at this point for non-physical spells. I'll look into that after this is done.

sim/core/spell_school.go
FelixPflaum: I did not touch this, it's only used for getting the school of the boss auto attacks if I see it right. The proto only has physical=0 to schadow=6, so this function covers all possible cases currently.

FelixPflaum:
@kayla-glick The only commit in here that broke test results was [0d0e589](https://github.com/wowsims/sod/pull/342/commits/0d0e5897fa4a6d4b145659d06341c9e0b163a137), which fixed level based resistance. The now slightly increased effective resistance against higher level targets causes slightly more partials across the board for everything that isn't physical damage.

Caught this because the tests I added check for the expected value of 0.08(*) for the resistance coef at +3 levels, the same the calc and guide at https://royalgiraffe.github.io/legacy-sim/#/resistances also get. I'm also checking for known values taken from there in the tests. I think resistance *should* be correct now.

(*) Every level above the caster adds 2% average mitigation, but the coef is before the multiplication with 0.75, so every level actually needs to add 0.02/0.75 to the coef, instead of the previous 0.02.

sim/core/spell_result.go
kayla-glick: Ok so you think this isn't a blocking issue?

sim/warlock/shadowflame.go
```diff
@@ -62,63 +41,63 @@ func (warlock *Warlock) registerShadowflameSpell() {
 			TickLength:    time.Second * 2,
 
 			OnSnapshot: func(sim *core.Simulation, target *core.Unit, dot *core.Dot, isRollover bool) {
-				// Use fire school for dot modifiers
-				dot.Spell.SpellSchool = core.SpellSchoolFire
-				dot.Spell.SchoolIndex = stats.SchoolIndexFire
-
 				dot.SnapshotBaseDamage = dotDamage + dotSpellCoeff*dot.Spell.SpellDamage()
 				dot.SnapshotBaseDamage *= emberstormMulti
```
rosenrusinov, line 45: this can be moved to the damage multi of the dot spell config

rosenrusinov (approved):

PR #344, opened 2024-03-04
Warrior stuff

vigo2:
[warrior] fix Quick Strike's rage cost
[warrior] fix Rend not dealing damage, having the wrong number of ticks, and scaling wrongly from Improved Rend ;) [warrior] non-instant Slam now resets the swing timer
[warrior] changing stances doesn't retain 10 rage by default anymore 
[warrior] Deep Wounds doesn't double dip from "damage taken" modifiers anymore, but double dips from more "damage done" modifiers 

[general] added mana/rage/energy costs to Berserking (and clamp "berserkingPct" between 0.1 and 0.3)

I'm not sure about the Slam (reset, not just delay the swing timer) and Stance Dancing (0 rage base, not 10 rage base retained) changes.

Vanilla Deep Wounds had a _very_ weird interaction with "BonusWeaponDamage" auras, I'm pretty sure that doesn't exist anymore.
The "damage taken" handling in Deep Wounds was specifically in place for Blood Frenzy and the "no double dipping"-WotLK policy, so I removed it as well. My guess is "damage taken" modifiers don't "double dip", but that could easily be tested duelling against a warrior in Defensive Stance.

### DIFF
--- a/sim/core/racials.go
+++ b/sim/core/racials.go
@@ -2,7 +2,6 @@ package core
 
 import (
 	"fmt"
-	"math"
 	"time"
 
 	"github.com/wowsims/sod/sim/core/proto"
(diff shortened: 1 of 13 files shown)


sim/core/racials.go
```diff
@@ -153,36 +152,41 @@ func makeBerserkingCooldown(character *Character, customPercentage float64, time
...
 		}
+		// from 10% at full health to 30% at 40% or less health
+		switch hp := character.CurrentHealthPercent(); {
+		case hp >= 1:
+			return 0.1
+		case hp <= 0.4:
+			return 0.3
+		default:
+			return 0.1 + (1-hp)/3
+		}
```
kayla-glick, line 166: ❓ Do the numbers change at all here? I'm trying to understand why the shaman tests changed

vigo2: Tests changed because of "added mana/rage/energy costs to Berserking". This here is the "clamp berserkingPct between 0.1 and 0.3 part", for the odd-ball case where Berserking is activated while e.g. Last Stand is active.

vigo2: So the numbers here haven't changed at all, since that latter scenario isn't tested (yet).

TheGroxEmpire (approved):

PR #363, opened 2024-03-05
Garrote fix

sanguinerarogue:
Updated garrote to calculate the damage per tick and used classic AP scaling


catszeid (approved):
Looks good to me

PR #364, opened 2024-03-05
Rogue stuff

vigo2:
[rogue] various smaller fixes and (hopefully ;) improvements:
- Ambush can no longer be dodged, blocked, or parried, and doesn't benefit from Lethality
- Eviscerate AP scaling reduced to 3% per CP
- Garrote AP scaling reduced to 3% per tick, and fixed base damage to per tick damage
- Ghostly Strike and Hemorrhage are using non-normalized damage now
- fixed Mutilate double dipping from Opportunity, and its interaction with Cold Blood
- lowered Improved Poison's proc chance bonus to 2% per talent point
- fixed Deadly Poison's damage calculation (each stack now using 0.9% instead of 3.5% AP contribution, and total tick damage multiplied by # of stacks)
- fixed Rupture ticks to no longer crit, and reduced its AP scaling (4/10/18/21/24% instead of 6/12/18/24/30%); also fixed base damage to per tick damage
- Sword Specialization now just grants extra attacks (and its ICD was lowered)
- Blade Flurry now double dips from damage multipliers (maybe wrong, but unlikely)


sim/rogue/garrote.go
```diff
@@ -51,10 +50,10 @@ func (rogue *Rogue) registerGarrote() {
 				Label: "Garrote",
 				Tag:   RogueBleedTag,
 			},
-			NumberOfTicks: numTicks,
+			NumberOfTicks: 6,
 			TickLength:    time.Second * 3,
 			OnSnapshot: func(sim *core.Simulation, target *core.Unit, dot *core.Dot, _ bool) {
-				dot.SnapshotBaseDamage = baseDamageTick + dot.Spell.MeleeAttackPower()*0.03
+				dot.SnapshotBaseDamage = baseDamage + dot.Spell.MeleeAttackPower()*0.07
```
catszeid, line 56: AP scaling should be adjusted to 3% as per testing results.

sim/rogue/talents.go
```diff
@@ -96,8 +96,8 @@ func (rogue *Rogue) registerColdBloodCD() {
 			}
 		},
 		OnSpellHitDealt: func(aura *core.Aura, sim *core.Simulation, spell *core.Spell, result *core.SpellResult) {
-			// for Fan of Knives and Mutilate, the offhand hit comes first and is ignored, so the aura doesn't fade too early
-			if spell.Flags.Matches(SpellFlagColdBlooded) && spell.ProcMask.Matches(core.ProcMaskMeleeMH) {
+			// for Mutilate, the mainhand hit comes first and is ignored, so the aura doesn't fade too early
+			if spell.Flags.Matches(SpellFlagColdBlooded) && spell.ProcMask.Matches(core.ProcMaskMeleeOH) {
```
catszeid, line 100: Conditional should be adjusted to include SpellID of Mutilate OH since there are now several offhand abilities that should be affected by Cold Blood. These include Shiv, Poisoned Knife, and Main Gauche.

catszeid (changes requested):
Looks good other than the 2 comments

sim/rogue/talents.go
vigo2: This is for disabling the aura after a "cold blooded" ability used it, so Shiv & friends would all still profit. But it's still wrong, since those spells now profit, but don't disable Cold Blood anymore.

catszeid: Mutilate Hit order could be reversed again if that would fix this issue.

catszeid (approved):

PR #372, opened 2024-03-08
Poisons / Windfury / Wild Strikes

vigo2:
[rogue] Main Gauche, Saber Slash, and Shiv now benefit from Aggression
[rogue] fix Shiv to work with Deadly Brew (applying IP); remove the +100% spell hit buff (the 100% is - traditionally - the "proc chance" for Shiv-applied poisons), and allow Shiv's IP to crit 
[rogue] introduce "SpellFlagPoison" for poisons, to simplify "Deadly Brew (Deadly)" aura [rogue] snapshot DP only when adding a stack
[rogue] likewise change Saber Slash's DOT

[druid] allow cat abilities to proc Wild Strikes; found log evidence for Mangle, Rake, Shred, and Rip, and simply assume FB works the same

[warrior, druid, hunter] next melee abilities now have ProcMaskMeleeMHSpecial | core.ProcMaskMeleeMHAuto, since they behave like autos in almost all circumstances

[core] changed RageBar proc masks accordingly

[warrior] Consumed By Rage now loses stacks on any melee hit, not just white hits 
[warrior] Unbridled Wrath doesn't use PPM anymore, and correctly (indirectly, see above) triggers from HS and Cleave now 
[warrior] righteously, 5/5 Flurry is back to 30% attack speed

[core] Wild Strikes now simply adds 20% of the current AP, instead of temporarily adding a 20% AP multiplier 
[core] Windfury Totem immediately loses a stack if triggered by a MH auto (for simplicity, it just starts at 1 stack in that case); it's decidedly worse than Wild Strikes now 
[core] both Wild Strikes' and Windfury Totem's charges now get removed whenever an auto or next melee attack happens during buff aura uptime, irrespective of main- or offhand, and irrespective of whether it lands or not


sim/rogue/rogue.go
```diff
@@ -9,8 +9,9 @@ import (
 )
 
 const (
-	SpellFlagBuilder     = core.SpellFlagAgentReserved2
-	SpellFlagFinisher    = core.SpellFlagAgentReserved3
+	SpellFlagBuilder     = core.SpellFlagAgentReserved1
+	SpellFlagFinisher    = core.SpellFlagAgentReserved2
+	SpellFlagPoison      = core.SpellFlagAgentReserved3
```
kayla-glick, line 14: We'd like to make a core poison spell flag so that we can apply the 20% poison DR correctly. Is that something that would work in place of this rogue-specific flag?

kayla-glick: It would be checked here https://github.com/wowsims/sod/blob/80f0d0f0db6d721b9baf25a21d2995c2cc6b2912/sim/core/spell_result.go#L518-L520 and afaik would apply to hunter's Serpent Sting and Viper Sting

vigo2: Semantically, it's more of a DeadlyBrew flag. If the `PoisonDamageTakenMultiplier` becomes a thing in SOD, there's neither much clash nor much synergy between this and a global flag.

sim/rogue/poisons.go
```diff
@@ -264,37 +264,26 @@ func (rogue *Rogue) makeInstantPoison(procSource PoisonProcSource) *core.Spell {
...
 		SpellSchool: core.SpellSchoolNature,
 		ProcMask:    core.ProcMaskWeaponProc,
+		Flags:       SpellFlagPoison,
 
 		DamageMultiplier: rogue.getPoisonDamageMultiplier(),
 		CritMultiplier:   rogue.SpellCritMultiplier(),
 		ThreatMultiplier: 1,
 
-		// Shiv gives 100% hit to always hit (except for 1% miss of magic table)
-		BonusHitRating: core.TernaryFloat64(isShivProc, 100*core.SpellHitRatingPerHitChance, 0),
```
catszeid, line 281: Based on testing against high level mobs, the application of an applied poison does not fail to apply other than the normal 1% to miss that can never be removed. The prime example of this is a level 12 bear fully resisting application of an applied poison which is only explained by magic hit table. Similarly, Poison Knife (using the same mechanic as Shiv) never failed to apply the applied poison (other than a very rare 1% or so) despite the target being level 50+. 

When testing with no poison applied and Deadly Brew engraved poison against a 50+ mob poison applications were frequently resisted and notably different in rate compared to when the same poison was applied to the weapon.

catszeid (commented):
Shiv is behaving differently in game compared to current implemention of your code with regards to application depending on whether it is an applied poison or triggered from Deadly Brew

TheGroxEmpire (approved):

sim/rogue/poisons.go
vigo2: I'll add this to all Shiv-Procced poisons then.

PR #402, opened 2024-03-15
Multi-School Refactoring

vigo2:
[core] simplify the multi school implementation (just switch on SpellSchool.Matches()); remove the various related constants (SpellStrike, Frostfire, Holymoly etc.), and remov the largely unknown mechanics for physical x magic resistances

[core] remove SchoolCritTakenMultiplier pseudo stat, since the only use case (Winter's Chill) is, in fact, not a use case for such a multiplier

I've looked at some logs for "Bloodbark Crusher" (`core.SpellSchoolPhysical | core.SpellSchoolNature`) and "Exsanguinar" (`core.SpellSchoolPhysical | core.SpellSchoolShadow`), and even these two work entirely different: the first seemed to behave mostly like `SpellSchoolNature`, the latter rather like a bleed, but the initial damage was "mostly random" (it should deal a fixed 65 damage, had rawDamage 0 (!), and varied from 50 to 150 or thereabouts, irrespective of Shadow or Physical damage modifiers present)

TLDR: It's utterly unclear how physical x magic spells work


FelixPflaum:
Physical multi school isn't "entirely unknown". Armor is used if armor value is ~resistance or less, only the specific rules for that are unknown because it wasn't tested much. Armor should not be used most of the time as a result. That's why I implemented it the way it is.

FelixPflaum:
But yeah, the switch to using indices was mostly due to performance considerations, and that became irrelevant in the end. That may be an ok change.

FelixPflaum:
Removing tests for multi school entirely is probably bad though. Resistance and bonus power for multi school still works differently and isn't tested anywhere else, right?

vigo2:
"physical + magic": 

If the "specific rules are unknown" at this point (and this includes how damage multipliers are used), the items using that mechanic are less than a handful, and the current implementation would choose the "magic" school basically anytime (resists typically being 0, and armor hardly ever below 400+), I'd either ignore those spells for now (that's what I'm doing), or change them to just use the "magic" school until there's a better theoretical foundation.

"testing":

I can add table tests for `unit.resistance()`, `spell.SpellSchoolPower()`, and `unit.GetSchoolDamageDoneMultiplier()`, but don't think they're overly necessary, given how simply these have become

FelixPflaum:
Tests are not necessary if nobody ever changes the code, sure, but e.g. you are doing exactly that. Breaking ms logic also doesn't necessarily break sim results, which is why I made explicit tests for it to begin with.

And as already discussed in discord, always using armor is just wrong. Always using resistance would still be more wrong than what I did. I just don't see a reason to make it more wrong just because it's not 100% correct the way it is.

kayla-glick:
Hey @FelixPflaum @vigo2 how are we feeling about this one?

FelixPflaum:
@kayla-glick I think removing all the individual indices is a good idea, they are simply not needed with how the implementation ended up, but I dislike other changes as mentioned above. Because of that I fiddled around with my own take at a refactor on top of some fixes like a week ago but got distracted (Elden Ring is a hell of a drug). Seeing as there were no changes here I opened a PR for that now.

Issue #377, opened 2024-03-09
Trickster's Helmet (Or generic green items) does not show Spell Damage suffixes for Rogue


Lisonna:
**Describe the bug**
Trickster's Helmet is filtering stats for random suffixes such that caster stats do not show for Rogue. 

**To Reproduce**
Find a green which has a Nature's Wrath suffix in-game or on wowhead, and search for it in the Sim, and look at available suffixes. 

![image](https://github.com/wowsims/sod/assets/70078996/6c087246-cece-4c0e-910d-8a20f52b0692)
This list does not show the above helmet option. It also does not show any suffixes with Intellect (though I don't think it should necessarily).
![image](https://github.com/wowsims/sod/assets/70078996/c5a8037a-88dd-470d-ad50-ee3ecc942a85)

**Expected behavior**
With Nature Damage benefiting Envenom, it would be reasonable to display Nature's Wrath as an available suffix on the helm (Or any item that has Nature's Wrath as a suffix).

kayla-glick:
Looks like we need to add nature power as an EP and display stat if we want this to be a thing. That would also require users to restore defaults in order to reset their EPs

Lisonna:
FWIW it is pretty easy to mock up this (or similar) item, by just unequipping the slot and adding a flat amount of spell damage in the stats table. Still overhead that unfamiliar users may not think to do and thus they are prevented from easily simming the option, but it isn't show-stopping and impossible to test without the item in the available set for Rogue.

kayla-glick:
This should be fixed by @sanguinerarogue's PR BUT the caveat is that EP weight updates like that require users to `Restore Defaults`

PR #409, opened 2024-03-17
Update rogue APL and tests to P2 presets

sanguinerarogue:
Added Mutilate IEA APL
Added AR/BF default talents and APL
Removed axes from rogue weapons
Added Nature Spellpower so it appears as an option on greens

Updated P2 tests to include P2 gear/talents/apl


sanguinerarogue:
Addresses #377

catszeid (approved):
Looks good to me

PR #449, opened 2024-03-23
Multischool fix/refactor

FelixPflaum:
- Fix multi school resistance for physical+x DoTs. If resistance is the "best" choice then they go against resistance even if they are explicitly a bleed.
- Actually implement school crit chance taken (Winter's Chill). It's now added flat crit chance.
- Refactor multi school code similar to https://github.com/wowsims/sod/pull/402

Unlike the mentioned PR this only does away with the pseudo school indices and special resistance handling. There's only one index for multi schools now, because that and generated base indices are enough to handle multi schools wherever needed.

The helper for setting a spell's school is removed, because **every** instance of changing the school during the sim is a (lazy) hack. Changed comments to make that clear. Also made the comment on MultiSchoolShouldUseArmor more specific to clarify why it's there and what exactly the known limitations are.


sim/core/spell.go
```diff
@@ -75,11 +75,10 @@ type Spell struct {
 	// The unit who will perform this spell.
 	Unit *Unit
 
-	SpellSchool       SpellSchool // Schoolmask of all schools this spell uses. Use Spell.SetSchool() to change this!
+	SpellSchool       SpellSchool         // Schoolmask of all schools this spell uses. Do not change this! Whatever you try to do is a hack and probably wrong.
+	SchoolIndex       stats.SchoolIndex   // Do not change this! Whatever you try to do is a hack and probably wrong.
+	SchoolBaseIndices []stats.SchoolIndex // Base school indices for multi schools. Do not change this! Whatever you try to do is a hack and probably wrong.
```
kayla-glick, line 80: Nice comment 😂

kayla-glick (approved):

PR #451, opened 2024-03-25
Rogue P3 Runes

catszeid:
Add confirmed and expected P3 runes for Rogue

Confirmed:
- Cut to the Chase
- Honor Among Thieves

Expected:
- Combat Potency
- Focused Attacks


sim/rogue/talents.go
```diff
@@ -299,11 +298,10 @@ func (rogue *Rogue) applyCombatPotency() {
 }
```
vigo2, line 298: Combat Potency has ProcMask 20, so it should be `ProcMaskMeleeOH = ProcMaskMeleeOHAuto | ProcMaskMeleeOHSpecial`.

sim/rogue/talents.go
```diff
@@ -316,16 +314,14 @@ func (rogue *Rogue) applyFocusedAttacks() {
 			if !spell.ProcMask.Matches(core.ProcMaskMelee) || !result.DidCrit() {
```
vigo2, line 314: Fan of Knives has ProcMask 340, so it should use `ProcMaskMeleeOrRanged`.

sim/rogue/talents.go
```diff
@@ -475,16 +471,14 @@ func (rogue *Rogue) lethality() float64 {
 	return 0.06 * float64(rogue.Talents.Lethality)
 }
 
-/** Honor Among Thieves (Possible P2 rune)
 func (rogue *Rogue) registerHonorAmongThieves() {
 	// When anyone in your group critically hits with a damage or healing spell or ability,
 	// you have a [33%/66%/100%] chance to gain a combo point on your current target.
 	// This effect cannot occur more than once per second.
-	if rogue.Talents.HonorAmongThieves == 0 {
+	if !rogue.HasRune(proto.RogueRune_RuneHonorAmongThieves) {
```
vigo2, line 478: nit: the comment here and below could be removed

vigo2 (approved):

PR #452, opened 2024-03-25
Shaman: Phase 3 runes and adjustments

kayla-glick:


sim/core/apl_actions_misc.go
```diff
@@ -45,6 +46,12 @@ type APLActionActivateAura struct {
 	aura *Aura
 }
 
+type APLActionActivateAuraWithStacks struct {
```
rosenrusinov, line 49: can you re-order the new code so the new APL actions methods are next to each other. Right now it looks like they are around other actions functions and its confusing

sim/priest/shadow_word_pain.go
```diff
@@ -90,17 +90,17 @@ func (priest *Priest) getShadowWordPainConfig(rank int) core.SpellConfig {
...
+				if result.Landed() {
+					spell.SpellMetrics[target.UnitIndex].Hits--
 					priest.AddShadowWeavingStack(sim, curTarget)
 					spell.Dot(curTarget).Apply(sim)
-					curTarget = sim.Environment.NextTargetUnit(curTarget)
+					spell.DealOutcome(sim, result)
 				}
+				curTarget = sim.Environment.NextTargetUnit(curTarget)
 			}
-			spell.DealOutcome(sim, result)
```
rosenrusinov, line 103: Did someone confirm this is how the chaning of the dot happens? I had asked before if it rolls separate hits and was told only the main cast rolls hit then every other chained dot is just applied with it but it could have been wrong. Just making sure this is an intended change

kayla-glick: I asked yesterday and was told by Tako in the priest discord that each one rolls. I'm also pretty sure I've seen that happen in game

kayla-glick: The only reason I changed this is that I was implementing Burn which sounds like it works the same way

Issue #447, opened 2024-03-23
Deadly Brew + Deadly Poison Interaction 


sanguinerarogue:
**Describe the bug**
When adding deadly poison while having deadly poison DPS climbs much higher than expected.

**To Reproduce**
Set rogue sim defaults, add deadly poison as a weapon enchant

**Sim Links and Screenshots**
[If applicable, add links to your sim settings or screenshots to help explain your problem.](https://wowsims.github.io/sod/rogue/#eJztVE9IFFEYn29mdv58u7M934rNviK2PU0TG7OrW2qHnTyFRXjoIN0sNkpMpSAqImwg2U4uHSKFQKIudSikoFurBCldCiyCEFbpIHURPIQU0Xuza4lQHTLq0Hf4+Pj9fu/7/ZjhPTQNOSX7clmGjoQNPlQBHsCifgKmARYBqMSum6h19fWcL54mEdtJae4IoEq+lC3eq4GFCrl/ADVydYf9UsMomQwse0Fz5h7aqJPhUbDfCPQZR0u6M8tRgzwvW87NR2KaDSznLceiZI4LPoJTWknytY9vCIvbgbBYWY7x/o7POrnHRXdb0STLtnctJkaVfAisXMKJtCr7lUHtnD4EdhmUCkD7sIpVWfpff7ToOPztCL+q97LUEfGhBNBJPC/f7DVnveZ8PpvJePkutds4KvXeUWwFE2j26hgxPlmOmtZQRdk7Q5tYIxI3bqhUviCxEG312iWaYTsxTVMYSyMaExBBhZ9br8zmGSfI5730EruIR2h3KACqBCAJBiGPSXczR2JTYFKZiDWlplXKSkfRnAC+yhDXQpzV1sZA41ZoPN+f24YmE9krcQdQZzUUDaE0+P2ijNm4ybVEuM669e5QVYn/mJvvpwPsFB6jPXXvtbm3uEmOxKcgGuaO8NyOXCOz6LFdP0++zm/Pt6BbGat/xe9mLSHLn47BaRgCGAMpkGUqpZ8oZu0PH/IbakOjnxwbFTVTaKshLwrbj4f1upAbUYzJNmocLJ4t9qVaPNt1/7H3YemyvyF7cPy39jyNNezbkBy8smH3R/y2kwNLV2YOvyqsrva74StgEpNe)

![image](https://github.com/wowsims/sod/assets/118020821/011e41dc-507e-4f98-b295-53482580cf31)

PR #454, opened 2024-03-26
DP fix

vigo2:
[rogue] use a common dot spell for all DP applications, so it doesn't matter how DP was applied

Fixes https://github.com/wowsims/sod/issues/447.

If applied via Deadly Brew or Shiv, DP applications and DP ticks can now be differentiated. Either the accounting isn't working as I'm expecting, or Deadly Brew somehow casts more DP applications than IP hits, which sounds wrong.

That's pre-existing, however, so not in this PR.


catszeid (approved):
Looks good to me. Make sure the sources are showing properly in the results breakdown

PR #466, opened 2024-03-27
Update Lava Lash for P3

Foknetics:


kayla-glick:
Hey @Foknetics I have this included in https://github.com/wowsims/sod/pull/452

PR #470, opened 2024-03-28
Implements battle squawk!

sanguinerarogue:
@kayla-glick If you don't mind reviewing this. I tested it to ensure it applies the aura at 0s, hasted for 4 minutes and removes after 4 minutes.  Checked the attack speed values for 1-5 chicken squawks.


proto/common.proto
```diff
@@ -518,6 +518,7 @@ message RaidBuffs {
 	TristateEffect stoneskin_totem = 24;
 	TristateEffect retribution_aura = 25;
 	bool sanctity_aura = 36;
+	int32 battle_squawk = 37;
```
kayla-glick, line 521: This is technically a party buff. A lot of our buffs are mis-labeled as raid buffs when it should be a party buff. It only _really_ matters for the raid sim and I don't think any of us have the time to maintain that for SoD so up to you if you want to change it

sanguinerarogue: Yea, a lot is hold over from raid wide in Wrath with LOTP/Auras etc all party in SoD.  I think its fine for you to merge as is, and if the raid sim every gets to fruition will need to remap every buff anyways.

Issue #478, opened 2024-03-28
Phase 3: Consumables


kayla-glick:
This issue is to track the state of consumables for phase 3 - e.g. what's available, what's not, and what are we missing and need to implement.

Still need new phase 3 consumables

### Flasks
Vanilla flasks can only be crafted in Scholomance and as such I suspect will be completely unavailable in phase 3, We should consider disabling them in the UI for now.
- [x] [Flask of Restless Dreams](https://www.wowhead.com/classic/item=222952/flask-of-restless-dreams)
  - New SoD alchemy-only flask
- [x] [Flask of Everlasting Nightmares](https://www.wowhead.com/classic/item=221024/flask-of-everlasting-nightmares)
  - New SoD alchemy-only flask

### Elixirs
- [x] [Elixir of the Mongoose](https://www.wowhead.com/classic/item=13452/elixir-of-the-mongoose)
  - Obtainable from a drop that should be reasonable
- [ ] [Mageblood Potion](https://www.wowhead.com/classic/item=20007/mageblood-potion)
  - Locked behind ZG rep so I think we can count this out?
- [x] [Elixir of Giants](https://www.wowhead.com/classic/item=9206/elixir-of-giants)
  - Obtainable from a drop that should be reasonable
- [ ] [Elixir of Superior Defense](https://www.wowhead.com/classic/item=13445/elixir-of-superior-defense)
  - Purchasable recipe from a major city
- [ ] [Strong Troll's Blood Potion](https://www.wowhead.com/classic/item=3388/strong-trolls-blood-potion)
  - Trainable
- [ ] [Elixir of Fortitude](https://www.wowhead.com/classic/item=3825/elixir-of-fortitude)
  - Trainable
- [x] [Greater Arcane Elixir](https://www.wowhead.com/classic/item=13454/greater-arcane-elixir)
  - Only realistic drop is from mobs in BRD at an incredibly low drop rate. Probably not obtainable to 99% of players if at all
- [x] [Elixir of Shadow Power](https://www.wowhead.com/classic/spell=11474/shadow-power)
  - Purchasable recipe from a major city
- [x] [Elixir of Greater Firepower](https://www.wowhead.com/classic/item=21546/elixir-of-greater-firepower)
  - Obtainable from mobs in Searing Gorge

### Weapon Imbues
- [x] [Brilliant Wizard Oil](https://www.wowhead.com/classic/item=20749/brilliant-wizard-oil)
  - Locked behind ZG rep so I think we can count this out?
[...]

PR #490, opened 2024-03-29
Phase 3 Flasks

kayla-glick:
Part of https://github.com/wowsims/sod/issues/478

Disables all vanilla flasks as they're crafted in Scholomance.

Adds the two new Alchemy-only flasks added in Phase 3


rosenrusinov:
Re-enables all vanilla flasks but adds a minLevel 51 to them.
Adds all new Atal'Ai potions for Sunken Temple

PR #494, opened 2024-03-29
Implemented Lock and Load helmet rune for hunters, added frost trap a…

ncberman:
Implementation of the hunter helmet rune `Lock and Load` from the [phase 3 announcement video](https://youtu.be/vDRuPJkGp_I?si=69m92E2KUeYq-r5U&t=389)

Implementation Details:
- Added new aura to hunter.go
- Added rune logic to hunter/runes.go
- Aura is activate after immolation trap, explosive trap, or frost trap are cast
- Aura is uses the OnCastComplete trigger and checks if the cast is a shot ability (determined using a combination of procmask and spell flag)
- When cast is determined to be a shot ability the aura is deactivated, mana is refunded, and spell cooldown is reset

Frost Trap has also been added as a spell! I thought this addition would be necessary since frost traps and fire traps do not share cooldowns and as per the reading of the rune should also trigger lock and load procs. The frost trap spell implementation does not do anything except for cost mana and trigger lock and load.


rosenrusinov (approved):

rosenrusinov:
Hey @ncberman i just merged this but i realized a small mistake. The mana refund part should ise spell.CurCast instead of default cast to make sure you are including any changes to the manacost that would have been applied. I can fix it later when im home or you can fix it with your next p3 rune implementation 😉

Issue #461, opened 2024-03-27
Phase 3 Rogue Sim TODO


sanguinerarogue:
Working Rogue Discord list of implementations and testing for phase three:

**Runes**

- [x] Carnage

      1. Which abilities are modified? (Builders, Finishers, Poisons, Procs)
      2. Does rupture/garrote gain the bonus from applying Carnage?
      3. Does Saber Slash gain the bonus from applying Carnage?
      4. Does second rogues Carnage apply to first rogues attacks?
 
- [x] Honor Among Thieves

      1. Do pet crits count?

**Items**

- [x] Enchanting Sigils - Does P1 and P2 items stack? https://www.wowhead.com/classic/spell=446243/sigil-of-living-dreams + https://www.wowhead.com/classic/spell=439156/sigil-of-innovation

vigo2:
Carnage is using "Apply Aura: Mod All Damage Done % by Caster" (like e.g. Haunt, Noxious Sting in WotLK), so:

1) Which abilities are modified? (Builders, Finishers, Poisons, Procs)
Per "affected spells" list in https://www.wowhead.com/classic/spell=432275/carnage, all rogue abilities are affected ("Is Ability"); this includes Builders, Finishers, Poisons, but auto attacks and item procs are _not_ affected

2. Does rupture/garrote gain the bonus from applying Carnage?
These aura behave like "Damage Taken" auras, so their damage multiplier is applied dynamically, and never snapshot (like e.g. Mangle). So both gain the bonus, for all ticks.

3. Does Saber Slash gain the bonus from applying Carnage?
The first Saber Slash's direct damage on a non-bleeding target is most likely not affected, since the bleed debuff is applied after the hit, as per logs.

4. Does second rogues Carnage apply to first rogues attacks?
No, these are "per caster".

PR #496, opened 2024-03-29
Carnage

vigo2:
[core] removed unused PseudoStats, AttackTable properties, and SpellFlags
[core] introduced support for "damage done by caster" target multipliers via callback, so spell filtering can be done in per-class code (used for Haunt and Carnage)

[rogue] refactored flags usage, and added SpellFlagCarnage to all abilities affected by Carnage; Poisoned Knife and Quick Draw are now affected by Cold Blood, too 
[rogue] added Carnage support (as theorycrafted in https://github.com/wowsims/sod/issues/461)

[warlock] introduce SpellFlagHaunt as agent spell flag, and use instead of global SpellFlagHauntSE


catszeid (commented):
Rogue related code looks good. Leaving in depth review of the rest for others

PR #498, opened 2024-03-29
Druid: Add phase 3 rune support

kayla-glick:


sim/core/apl_values_aura.go
```diff
@@ -25,7 +22,7 @@ func (value *APLValueAuraIsActive) Type() proto.APLValueType {
 	return proto.APLValueType_ValueTypeBool
 }
 func (value *APLValueAuraIsActive) GetBool(sim *Simulation) bool {
-	return value.aura.Get().IsActive()
+	return value.aura.Get() != nil && value.aura.Get().IsActive()
```
kayla-glick, line 25: This fixes an issue where if you have an APL action with a condition on an aura that isn't available (e.g. rune not equipped, talent not taken, etc) the condition was evaluating to true and the action would be performed. In reality, the condition should evaluate to false and the action not taken


Example APL action for this
image

sim/druid/runes.go
```diff
@@ -9,33 +9,80 @@ import (
...
+func (druid *Druid) rollGoreBearReset(sim *core.Simulation) {
+	if sim.RandomFloat("Gore (Bear)") < Gore_BearResetProcChance {
+		druid.MangleBear.CD.Reset()
+	}
+}
+
+func (druid *Druid) rollGoreCatReset(sim *core.Simulation) {
+	if sim.RandomFloat("Gore (Cat)") < Gore_CatResetProcChance {
+		druid.TigersFury.CD.Reset()
+	}
```
kayla-glick, line 85: I opted to break up Gore and Elune's Fires like this for performance reasons. Originally I gave each one an aura with on `OnSpellHitDealt` with different conditions but that adds a lot of extra conditionals on every single hit and ultimately I think this approach requires a bit more dev work but will be more performant since the relevant code is only being checked when the actual trigger spells hit

ui/core/components/detailed_results/timeline.tsx
```diff
@@ -866,6 +866,11 @@ export class Timeline extends ResultComponent {
 				});
 		});
 
+		// If there are any auras that correspond to this cast, visualize them in the same row.
+		aurasById
+			.filter(auraUptimeLogs => auraUptimeLogs[0].actionId!.equalsIgnoringTag(actionId))
+			.forEach(auraUptimeLogs => this.applyAuraUptimeLogsToRow(auraUptimeLogs, rowElem));
```
kayla-glick, line 872: Added this back after removing it recently because it really helps with seeing dot duration

sim/core/apl_values_aura.go
rosenrusinov: I dont agree with this change. Every apl value we have is treated as not existing when there is a warning in it (just like here youre trying to use a nonexistent aura). This would mean that this single value becomes special and does not behave like every other value. I dont like this inconsistency. Instead we can add a new apl value for “Has Rune” if we want to make handling missing runes better in apl.

ui/core/components/detailed_results/timeline.tsx
rosenrusinov: I was actually gonna ask why was this broken😅
Was not aware you removed it on purpose and i think it should stay as well as you say it helps a lot with durations visibility

PR #502, opened 2024-03-30
Rune ability scaling refactor + Warrior fixes

vigo2:
[various] introduce and use runeAbility() to encapsulate the rune scaling calculations for each class

[warrior] simplify and fix the various rune abilities with weapon or shield requirements [warrior] fix Gladiator Stance SpellID
[warrior] add basic Shield Mastery support

[core] fix how PseudoStats.MeleeCritMultiplier applies in CritMultiplier()


TheGroxEmpire (approved):
Warrior's look good to me.

sim/druid/druid.go
```diff
@@ -290,6 +290,10 @@ func (druid *Druid) HasRune(rune proto.DruidRune) bool {
 	return druid.HasRuneById(int32(rune))
 }
 
+func (druid *Druid) runeAbility() float64 {
+	return 9.183105 + 0.616405*float64(druid.Level) + 0.028608*float64(druid.Level*druid.Level)
+}
```
kayla-glick, line 295: I'm really wary of this change. While this is true for now that we know of, Blizzard could easily add a new ability that uses a slightly different scaling equation and suddenly we can't rely on this as a "unified" equation

kayla-glick: If we do go this route can we rename the method to something that implies it's a damage value? `baseRuneAbilityDamage()`

vigo2: I'm fine with a rename, since `runeAbility` isn't great to start with. It's short, so that's something. But changing the name can be done in a simple refactor later, and might be a matter of each agent's taste.

kayla-glick: > I'm fine with a rename, since `runeAbility` isn't great to start with. It's short, so that's something. But changing the name can be done in a simple refactor later, and might be a matter of each agent's taste.

If the point is to create consistency by adding this method then we should keep it consistent across sims too

kayla-glick: Most other shared high-level functions are defined on the [Agent Interface](https://github.com/wowsims/sod/blob/1a8bdbd27256b644bd55f8e0241459e351ae93ba/sim/core/agent.go#L13-L46). Personally I would do the same here to enforce it across sims. A name refactor is also a simple mass find/replace so there's no reason to punt on it imo

vigo2: This was just meant as a little refactor, seeing some classes have these calculations repeat for a lot of spells.

It's not important enough to land in a core interface, and enforcing its implementation (especially if some agents don't need it at all, and some have classes have more than one).

PR #508, opened 2024-03-30
Implemented Hunter Focus Fire Rune

ncberman:
Implemented `Focus Fire` bracers rune for hunters as previewed in the [Phase 3 announcement video](https://youtu.be/vDRuPJkGp_I?si=x_lH2AApsiszBqvO&t=404)

Implementation Details:
- Created Aura for **hunters** labeled `Focus Fire` increases ranged attack speed by (1 * ( 0.03 * petFrenzyStacks)) when activated
- Created Aura for **hunter pets** labeled `Focus Fire Frenzy` which modifies pet attack speed whenever number of stacks change
- Created Aura for **hunters** labeled `Focus Fire Frenzy (Hunter)` that follows the pet's frenzy stacks so it can be tracked in the timeline
- Created Aura for **hunter pets** labeled `Focus Fire Pet` which is always active and tracks spell hits adding stacks to `Focus Fire Frenzy` and `Focus Fire Frenzy (Hunter)` whenever the pet uses an ability with ProcMaskMeleeMHAuto (should just be auto attacks)
- New Spell added for **hunter** called `Focus Fire` when used it activates `Focus Fire` aura gives the pet focus equal to 
(4 * petFrenzyStacks) and sets `Focus Fire Frenzy` stacks to 0.

Some values are currently unknown as it relates to the actual spell itself so I used cataclysm cooldown for the placeholder cooldown value (15 sec) and set mana cost to 0 since hunters use focus in cataclysm.


sim/hunter/focus_fire.go
```diff
@@ -0,0 +1,101 @@
...
+	"github.com/wowsims/sod/sim/core"
+	"github.com/wowsims/sod/sim/core/proto"
+)
+
+func (hunter *Hunter) registerFocusFireSpell() {
+	if hunter.pet == nil || !hunter.HasRune(proto.HunterRune_RuneBracersFocusFire) {
+		return
+	}
+
+	focusFireMetrics := hunter.pet.Metrics.NewResourceMetrics(core.ActionID{SpellID: 428726}, proto.ResourceType_ResourceTypeEnergy)
```
rosenrusinov, line 15: ResourceTypeFocus

sim/hunter/focus_fire.go
```diff
@@ -0,0 +1,101 @@
...
+		ActionID: focusFireActionId,
+		Duration: time.Second * 20,
+		OnGain: func(aura *core.Aura, sim *core.Simulation) {
+			frenzyStacks := hunterPetFrenzyAura.GetStacks()
+			aura.Unit.MultiplyRangedSpeed(sim, 1+(0.03*float64(frenzyStacks)))
+		},
+		OnExpire: func(aura *core.Aura, sim *core.Simulation) {
+			frenzyStacks := hunterPetFrenzyAura.GetStacks()
+			aura.Unit.MultiplyRangedSpeed(sim, 1/(1+(0.03*float64(frenzyStacks))))
+		},
```
rosenrusinov, line 54: the stacks on the pet aura can change between the OnGain/OnExpire (example OnGain with 1 stack, OnExpire with 5) which make stats "leak". You should setup this aura with its own stacks and use OnStacksChange the same way you do it for the pet frenzy aura. Then all you need to do is set the stacks to the current amount the pet frenzy has and the OnStacksChange will properly take care of the rest.

sim/hunter/focus_fire.go
```diff
@@ -0,0 +1,101 @@
...
+				Duration: time.Second * 15, 	// TODO: Update when cooldown is known (Currently using the cooldown value from Cataclysm)
+			},
+		},
+		ExtraCastCondition: func(sim *core.Simulation, target *core.Unit) bool {
+			return hunter.pet.IsEnabled() && (hunterPetFrenzyAura.GetStacks() > 0)
+		},
+		ApplyEffects: func(sim *core.Simulation, _ *core.Unit, _ *core.Spell) {
+			frenzyStacks := hunterPetFrenzyAura.GetStacks()
+			hunter.pet.AddFocus(sim, float64(4 * frenzyStacks), focusFireMetrics)
+			hunterFocusFireAura.Activate(sim)
```
rosenrusinov, line 97: `hunter.FocusFireAura.SetStacks(frenzyStacks)`

sim/hunter/focus_fire.go
ncberman: Updated in most recent commit

sim/hunter/focus_fire.go
```diff
@@ -0,0 +1,100 @@
...
+			},
+		},
+		ExtraCastCondition: func(sim *core.Simulation, target *core.Unit) bool {
+			return hunter.pet.IsEnabled() && (hunterPetFrenzyAura.GetStacks() > 0)
+		},
+		ApplyEffects: func(sim *core.Simulation, _ *core.Unit, _ *core.Spell) {
+			frenzyStacks := hunterPetFrenzyAura.GetStacks()
+			hunter.pet.AddFocus(sim, float64(4 * frenzyStacks), focusFireMetrics)
+
+			if !hunterFocusFireAura.IsActive() { hunterFocusFireAura.Activate(sim) }
```
rosenrusinov, line 95: You should not check if its active here because by doing so you wont refresh its duration if it was already active

sim/hunter/focus_fire.go
ncberman: I totally missed this, I updated to use your solution passing in the stack amount from the spell itself

sim/hunter/focus_fire.go
rosenrusinov: Just calling activate will properly refresh the aura if it was already active

ncberman: I'll make this change, but in the future should I always call activate an aura before changing stacks if I want the buff to refresh or should I call aura.Refresh inside of the OnStackChange hook?

rosenrusinov: Calling activate does not do anything to the stacks. It activates the aura if its not active or refreshes it if it is. Stack handling is done separately as auras have different rules for how their stacks work. In your case you want to always set stacks to the correct number after the activate

sim/hunter/focus_fire.go
```diff
@@ -0,0 +1,100 @@
...
+	"github.com/wowsims/sod/sim/core"
+	"github.com/wowsims/sod/sim/core/proto"
+)
+
+func (hunter *Hunter) registerFocusFireSpell() {
+	if hunter.pet == nil || !hunter.HasRune(proto.HunterRune_RuneBracersFocusFire) {
+		return
+	}
+
+	focusFireMetrics := hunter.pet.Metrics.NewResourceMetrics(core.ActionID{SpellID: 428726}, proto.ResourceType_ResourceTypeFocus)
```
rosenrusinov, line 15: just realized this should be `NewFocusMetrics(core.ActionID{})`

Issue #516, opened 2024-03-30
Phase 3: Implement new crafted armor effects


kayla-glick:
Phase 3 brings a number of new crafted armor pieces with special effects that need to be implemented. Thankfully Blizzard has been kind and gives us the proc chances / PPM / duration etc

PR #519, opened 2024-03-31
P3 crafted shoulder item effects

tyler-cb:
https://www.wowhead.com/classic/news/new-phase-3-profession-items-coming-to-classic-season-of-discovery-338284


kayla-glick:
https://github.com/wowsims/sod/issues/516

kayla-glick (approved):
We could break these `Echoes of X` auras out into functions in this file to save on repetition since a lot of them are shared by multiple items but I wouldn't consider that a blocker. Just means a little less code and less chance for typos 😄

sim/common/sod/crafted/phase_3.go
```diff
@@ -0,0 +1,341 @@
...
+		})
+
+		handler := func(sim *core.Simulation, _ *core.Spell, _ *core.SpellResult) {
+			procAura.Activate(sim)
+		}
+
+		core.MakeProcTriggerAura(&character.Unit, core.ProcTrigger{
+			ActionID:   core.ActionID{SpellID: 446579},
+			Name:       "Echoes of Dread",
+			Callback:   core.CallbackOnCastComplete,
```
rosenrusinov, line 150: core.CallbackOnSpellHitDealt

rosenrusinov: Actually just checked wowhead and im wrong. It  should be cast complete

rosenrusinov (approved):

PR #522, opened 2024-03-31
Added missing shoulder effect + cleanup

tyler-cb:
Added https://www.wowhead.com/classic/item=220741/screaming-chain-pauldrons which I missed as the wowhead article didn't include it

### DIFF
--- a/sim/common/sod/crafted/phase_3.go
+++ b/sim/common/sod/crafted/phase_3.go
@@ -10,12 +10,7 @@ import (
 func init() {
 	core.AddEffectsToTest = false
 
-	///////////////////////////////////////////////////////////////////////////
-	//                                 Cloth
-	///////////////////////////////////////////////////////////////////////////
-
-	// Mantle of Insanity
-	core.NewItemEffect(220749, func(agent core.Agent) {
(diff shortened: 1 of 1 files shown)


sim/common/sod/crafted/phase_3.go
```diff
@@ -10,12 +10,7 @@ import (
 func init() {
 	core.AddEffectsToTest = false
 
-	///////////////////////////////////////////////////////////////////////////
-	//                                 Cloth
-	///////////////////////////////////////////////////////////////////////////
-
-	// Mantle of Insanity
-	core.NewItemEffect(220749, func(agent core.Agent) {
+	echoesOfInsanityEffect := func(agent core.Agent) {
```
kayla-glick, line 13: 🔨 We should probably go with top-level functions here. Apparently the Go compiler struggles a bit with nested functions and top-level are more performant

tyler-cb: Oh I see thanks, I'm new to go so i'll probably miss a lot of things like that

kayla-glick (approved):
Love it! :shipit:

PR #523, opened 2024-03-31
Implementation of Raptor Fury rune for hunter

ncberman:
Implementation of the Raptor Fury Rune from [wowhead datamining](https://www.wowhead.com/classic/spell=415358/raptor-fury)

Implementation Details:
- Created a new aura to track stacks of the raptor fury buff
- Raptor strike applies damage multiplier after flanking strike damage is applied but before the spell is cast
- Raptor strike adds a stack to the raptor fury buff after the spell is cast and does not refresh the buff duration

_I'm still new to adding runes to the sim so let me know if I'm ever overlooking patterns or methods that I should be using instead!_


sim/hunter/runes.go
```diff
@@ -199,4 +200,17 @@ func (hunter *Hunter) applyLockAndLoad() {
...
+}
+
+func (hunter *Hunter) applyRaptorFury() {
+	if !hunter.HasRune(proto.HunterRune_RuneBracersRaptorFury){
+		return
+	}
+
+	hunter.RaptorFuryAura = hunter.GetOrRegisterAura(core.Aura{
+		Label:     "Raptor Fury Buff",
+		ActionID:  core.ActionID{SpellID: 415358},
```
kayla-glick, line 212: 🔨 Tiny knit but you can pass `int32(proto.HunterRune_RuneBracersRaptorFury)` here in place of the ID because they're the same value 😄

sim/hunter/raptor_strike.go
```diff
@@ -122,6 +134,11 @@ func (hunter *Hunter) getRaptorStrikeConfig(rank int) core.SpellConfig {
 			if hasMeleeSpecialist && sim.RandomFloat("Raptor Strike Reset") < 0.3 {
 				spell.CD.Reset()
 			}
+
+			if hasRaptorFury {
+				if !hunter.RaptorFuryAura.IsActive() { hunter.RaptorFuryAura.Activate(sim) }
```
kayla-glick, line 139: 🔨 Tiny Knit: It looks like the last `}` didn't get pushed to the next line

kayla-glick: I think technically you don't have to check `IsActive()` here either and can just call 
```
hunter.RaptorFuryAura.Activate(sim)
hunter.RaptorFuryAura.AddStack(sim)
```

back-to-back

kayla-glick (commented):
Nothing major from me but I'll let @rosenrusinov review since he typically handles the hunter shenanigans. Thanks for contributing!

sim/hunter/raptor_strike.go
rosenrusinov: Actually the way he has it is correct. The ability states that it does not refresh its duration but keeps on stacking. So this is correct. But I do agree I dont like the formatting. Make sure you're following the styling options of the project which means new line on every curly bracket

sim/hunter/raptor_strike.go
```diff
@@ -85,13 +89,17 @@ func (hunter *Hunter) getRaptorStrikeConfig(rank int) core.SpellConfig {
 				spell.BonusWeaponDamage()
 
 			if hasFlankingStrike && hunter.FlankingStrikeAura.IsActive() {
-				mhBaseDamage *= 1.0 + (0.1 * float64(hunter.FlankingStrikeAura.GetStacks()))
+				mhBaseDamage *= 1.0 + (flankingStrikeDmgMult * float64(hunter.FlankingStrikeAura.GetStacks()))
```
rosenrusinov, line 92: set the flanking and raptor fury stacks into variables if you're using them in multiple lines. No point in calling `float64(GetStacks())` multiple times

rosenrusinov (commented):
Minor comments otherwise implementation looks correct

sim/hunter/raptor_strike.go
kayla-glick: > Actually the way he has it is correct. The ability states that it does not refresh its duration but keeps on stacking. So this is correct.

Ah good catch!

sim/hunter/raptor_strike.go
kayla-glick: This is totally not needed for this PR but I noticed most of the conditions are checked for both MH and OH e.g. `hasFlankingStrike && hunter.FlankingStrikeAura.IsActive()` we should maybe do the same with those

sim/hunter/runes.go
ncberman: all in favor of cleaning up my code! I'll look at using your suggestion

sim/hunter/raptor_strike.go
ncberman: I'll look at cleaning this stuff up with this PR since I'm going to be making one more commit with minor changes anyway

sim/hunter/raptor_strike.go
ncberman: Yeah I thought this was funny because my other PR I had the opposite problem where I wasn't refreshing when I should've been, I'll also include the formatting changes

PR #524, opened 2024-03-31
APL Core: Don't run actions when the provided condition evaluates to nil

kayla-glick:
This is a quirk I've noticed with the APL that shows up quite a bit when rune swapping and setting certain aura conditions. In these scenarios where a spell or aura isn't known, the `APLValue` of a condition using that spell/aura ends up being returned as `nil`, which in-turn evaluates to a `nil` condition, aka treated as if there was no condition at all.

### Single-Conditionals

In my mind, when I don't have the Omen of Clarity talent taken (allow clearcasting procs, this condition below should not allow the action to run. The reason it does is that the condition evaluates to nil, so the APL treats it as if it had no condition at all. I don't think it's intuitive that the sim would cast Wrath here.

image

### Any-Ofs

`Any of`s where all conditions (regardless of count) use unavailable spells/auras suffer from the same problem as above and are fixed by this change. In the below screenshot, when I have not taken Omen of Clarity I would expect the sim to not cast Wrath. If, however, at least one condition is still able to be evaluated, then the sim treats the `Any of` the same both before and after these changes - only evaluating the one or more valid conditions.

image

### All-Ofs

`All of`s are a bit more complicated. Right now the sim treats them the same way as `Any of`s, where an `All of` with only condition using unknown spells/auras will evaluate to `nil` and perform the action. When adding additional conditions, only those conditions are evaluated.

The same changes to `Any of` apply to `All of`, e.g. the below condition will not allow the action to be run.

image

### Other Issues

Where I'm not sure about though is in a scenario like the one below. Right now in the sim, when I do not have the Omen of Clarity Talent this condition will still evaluate to `true` and the sim will cast Wrath. This remains true after these changes as well. This feels incorrect, as my assumption would be that this condition should evaluate to `false` because it's written to assume that all conditions need to be met regardless

image

### DIFF
--- a/sim/core/apl_action.go
+++ b/sim/core/apl_action.go
@@ -120,7 +120,7 @@ func (rot *APLRotation) newAPLAction(config *proto.APLAction) *APLAction {
 		impl:      rot.newAPLActionImpl(config),
 	}
 
-	if action.impl == nil {
+	if action.impl == nil || (config.Condition != nil && action.condition == nil) {
 		return nil
 	} else {
 		return action
--- a/sim/core/apl_values_aura.go
+++ b/sim/core/apl_values_aura.go
@@ -14,6 +14,9 @@ type APLValueAuraIsActive struct {
 
 func (rot *APLRotation) newValueAuraIsActive(config *proto.APLValueAuraIsActive) APLValue {
 	aura := rot.GetAPLAura(rot.GetSourceUnit(config.SourceUnit), config.AuraId)
+	if aura.Get() == nil {
+		return nil
+	}
 	return &APLValueAuraIsActive{
 		aura: aura,
 	}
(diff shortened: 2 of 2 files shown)


kayla-glick:
Looks like this breaks Ele. Arms, and Fury's APLs.

For Ele, this condition goes from evaluating to true to false when swapping off the mail helm, meaning other cooldowns are never cast.

image

Ideally I would want to be able to instead write:
```
if any of | Not Aura Known Hyperconductive Shock
              | Not Aura Active Hypercondutive Shock
```

This would require us to add an `Aura Known` value which I honestly think would be very useful

kayla-glick:
In the Fury sim there are two problem conditions:

image
image

Both of these conditions evaluate to `true` on the live sim and cast their respective abilities, but at face value I would expect them to evaluate to `false`

rosenrusinov:
Let's have a bigged discussion around this when we're both awake as I think its too big of a change and also the new proposed change feels wrong to me. (You are making values that dont compile actually return a value which just sounds wrong to me)

PR #533, opened 2024-03-31
Add new OH weapon debuff

Foknetics:
Adds new OH Fist debuff and weapon proc
[https://www.wowhead.com/classic/item=220589/serpents-striker](https://www.wowhead.com/classic/item=220589/serpents-striker)

### DIFF
--- a/sim/common/sod/item_effects/phase_3.go
+++ b/sim/common/sod/item_effects/phase_3.go
@@ -3,6 +3,7 @@ package item_effects
 import (
 	"time"
 
+	"github.com/wowsims/sod/sim/common/itemhelpers"
 	"github.com/wowsims/sod/sim/core"
 	"github.com/wowsims/sod/sim/core/stats"
 )
@@ -10,6 +11,7 @@ import (
 const (
 	AtalaiBloodRitualCharm = 220634
 	CobraFangClaw          = 220588
+	SerpentsStriker        = 220589
 )
 
 func init() {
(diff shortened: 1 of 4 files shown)


sim/common/sod/item_effects/phase_3.go
```diff
@@ -117,5 +119,25 @@ func init() {
 		})
 	})
 
+	// Serpent's Striker
+	itemhelpers.CreateWeaponProcSpell(SerpentsStriker, "Serpent's Striker", 5.0, func(character *core.Character) *core.Spell {
```
kayla-glick, line 123: Where did you find the 5% proc chance? Just want to verify it and I didn't see it on Wowhead

sim/common/sod/item_effects/phase_3.go
```diff
@@ -117,5 +119,25 @@ func init() {
...
+		procAuras := character.NewEnemyAuraArray(core.SerpentsStrikerFistDebuffAura)
+
+		return character.RegisterSpell(core.SpellConfig{
+			ActionID:    core.ActionID{SpellID: 447894},
+			SpellSchool: core.SpellSchoolNature,
+			DefenseType: core.DefenseTypeMagic,
+			ProcMask:    core.ProcMaskEmpty,
```
kayla-glick, line 130: We should add the `core.SpellFlagPoison` flag to this so that it's registered as a poison

sim/common/sod/item_effects/phase_3.go
Foknetics: I didn't, I don't think we have any way to know it until the item is live but I want to get it into the sim so this is modelled after Mekkatorque which is mechanically a very similar item.

kayla-glick: Ok that works for me for now. I'll leave a note to double check the proc chance

kayla-glick (approved):
:shipit:

PR #534, opened 2024-03-31
Phase 3 Gear Effects

kayla-glick:
### Bugs/Issues
Adding the [Serpent's Ascension](https://www.wowhead.com/classic/item-set=1651/serpents-ascension) set bonus caused problems with the test generation, because the offhand's aura was limited to level 50+ but tests were generating all tests regardless of level. I added a `RequiresLevel` field to DB items and a `Level` field on the test `ItemFilters` so that we can filter out any items with an invalid level requirement from the test generation

### Items Added
- [x] [Serpent's Ascension](https://www.wowhead.com/classic/item-set=1651/serpents-ascension) set bonus
- [x] [Ritualist's Hammer](https://www.wowhead.com/classic/item=221446/ritualists-hammer) (Druid Only)
- [x] [Gurubashi Pit Fighter's Bow](https://www.wowhead.com/classic/item=221450/gurubashi-pit-fighters-bow) (Hunter Only)
  - Also changed Bloodlash Bow from ppm to the 5% proc chance listed on Wowhead


sim/core/aura.go
```diff
@@ -910,11 +910,13 @@ func (caster *Unit) NewPartyAuraArray(makeAura func(*Unit) *Aura) AuraArray {
...
-		if target.Type == EnemyUnit {
-			auras[target.UnitIndex] = makeAura(target, caster.Level)
-		}
+	enemyUnits := FilterSlice(caster.Env.AllUnits, func(unit *Unit) bool {
+		return unit.Type == EnemyUnit
+	})
+
+	auras := make([]*Aura, len(enemyUnits))
+	for _, target := range enemyUnits {
+		auras[target.UnitIndex] = makeAura(target, caster.Level)
```
kayla-glick, line 919: This change limits the size of the aura array to only enemy units. This should be a bit more memory-performant as now the sim character and pets are not included

sim/core/test_generators.go
```diff
@@ -458,6 +462,10 @@ func FullCharacterTestSuiteGenerator(configs []CharacterSuiteConfig) []TestGener
 			defaultRaid.TargetDummies = 1
 		}
 
+		// Ensure we don't generate tests where the agent equips items above its level
+		// This previously caused bugs with effects with a specified minimum level above the agent's level
+		config.ItemFilter.Level = config.Level
```
kayla-glick, line 467: Filter out any gear above the test's level

sim/hunter/items.go
```diff
@@ -7,33 +7,39 @@ import (
...
+		hunter.newBloodlashProcItem(75)
+	})
+}
 
+func (hunter *Hunter) newBloodlashProcItem(bonusStrength float64) {
+	procAura := hunter.NewTemporaryStatsAura("Bloodlash", core.ActionID{SpellID: 436471}, stats.Stats{stats.Strength: bonusStrength}, time.Second*15)
+	core.MakePermanent(hunter.GetOrRegisterAura(core.Aura{
+		Label: "Bloodlash Trigger",
+		OnSpellHitDealt: func(aura *core.Aura, sim *core.Simulation, spell *core.Spell, result *core.SpellResult) {
+			if result.Landed() && sim.RandomFloat("Bloodlash") < .05 {
```
kayla-glick, line 40: Only noteworthy change is Bloodlash changed from ppm to 5% proc chance per the tooltips of both bows

sim/core/aura.go
rosenrusinov: Enemies are registered with their unit index so now the array size smaller then the last enemy index (due to player characters/pets). I actually wonder how is this working? Are enemies registered first so their unit indices start from 0 and it still works? Non the less i think the reason this was initialized with the full env units lenght is because we use unitIndex for indexing

sim/hunter/items.go
rosenrusinov: Im not sure about this change. This proc was tested by multiple people in hunter discord and they reported back 1ppm. Also most weapon procs are modeled with ppm and we know tooltip proc chances are sometimes wrong

sim/core/aura.go
kayla-glick: Ooooh good point. I totally overlooked that. I'll change it back

sim/hunter/items.go
kayla-glick: Ok I can also change that back then

Issue #532, opened 2024-03-31
Core: Add new APL "Spell Known" and "Aura Known" values


kayla-glick:
Follow-up to https://github.com/wowsims/sod/pull/524

APL conditions are incorrectly evaluated to true when a spell or aura is included in the condition that isn't known. This is because the missing spell/aura cause the condition to evaluate to `nil` instead of `false`. This allows some actions to be done when they shouldn't be. Adding these values would give us more granular control over creating actions that are only performed when a specific rune is enabled and condition met.

PR #536, opened 2024-04-01
Core: Add APL "Spell Known" and "Aura Known" values

kayla-glick:
Resolves https://github.com/wowsims/sod/issues/532

Right now many APLs are mistakenly casting abilities while using rune conditions when a rune isn't equipped because the conditions are evaluated to `nil`. This causes oddities when building around APL conditions for runes and sometimes talents or even gear that has proc auras (https://github.com/wowsims/sod/pull/524).

These new values allow us to ensure that we can write APL actions that are only performed when a certain rune is equipped, talent is known, item is equipped, etc.

I also updated some things for Feral to go along with this

### DIFF
--- a/sim/core/apl_values_aura.go
+++ b/sim/core/apl_values_aura.go
@@ -7,6 +7,27 @@ import (
 	"github.com/wowsims/sod/sim/core/proto"
 )
 
+type APLValueAuraIsKnown struct {
+	DefaultAPLValueImpl
+	aura AuraReference
+}
+
+func (rot *APLRotation) newValueAuraIsKnown(config *proto.APLValueAuraIsKnown) APLValue {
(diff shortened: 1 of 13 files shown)


sim/core/apl_values_aura.go
```diff
@@ -7,13 +7,39 @@ import (
...
 	DefaultAPLValueImpl
 	aura AuraReference
 }
 
 func (rot *APLRotation) newValueAuraIsActive(config *proto.APLValueAuraIsActive) APLValue {
 	aura := rot.GetAPLAura(rot.GetSourceUnit(config.SourceUnit), config.AuraId)
+
+	if aura.Get() == nil {
+		return nil
+	}
```
kayla-glick, line 41: These two additions are reverting a previous change. This brings `AuraIsActive` back to how it exists in Wrath and Cata as well

rosenrusinov (approved):

PR #544, opened 2024-04-01
implementation of catlike reflexes helm rune for hunter

ncberman:
Implementation of Catlike Reflexes rune from [wowhead datamining](https://www.wowhead.com/classic/spell=416083/engrave-helm-catlike-reflexes)

Details:
- Added a CD modifier variable for Kill Command & Flanking Strike
- When the hunter has this rune equipped the CD modifier on affected abilities will be halved (from 1.0 to 0.5)
- Added a stat change in hunter/runes.go when this rune is equipped to increase player and pet dodge change by 20% and 9% respectively.

_I'm still new to adding runes to the sim so let me know if I'm ever overlooking patterns or methods that I should be using instead!_


rosenrusinov (approved):
Looks good to me 👌

PR #551, opened 2024-04-02
Paladin P3

vigo2:
[paladin] add Hammer of Wrath, Holy Wrath, and Hammer of the Righteous
[paladin] support Fanaticism, Wrath, Improved Hammer of Wrath, and Purifying Power runes

Adapting to "house style" etc. can probably be done after this is merged. I've left out spell level scaling for Holy Wrath, because that's needlessly complicated for a one-time change at level 60. (Plus, half a ton of downranked spells are wrong, or were wrong in the phase they applied ;))


sim/paladin/talents.go
```diff
@@ -49,7 +49,7 @@ func (paladin *Paladin) getIlluminationActionID() core.ActionID {
 	}
 }
 
-func (paladin *Paladin) getBonusCritChanceFromHolyPower() float64 {
+func (paladin *Paladin) holyPower() float64 {
 	return core.CritRatingPerCritChance * float64(paladin.Talents.HolyPower)
 }
```
kayla-glick, line 54: 🚫 This change takes information away about what this method is doing. Longer, more descriptive method names are better than shorter ones that don't convey the proper intention

sim/paladin/runes.go
```diff
@@ -9,14 +9,24 @@ import (
...
+	// "RuneHeadFanaticism" is handled in Exorcism, Holy Shock, SoC, and SoR
+	// "RuneHeadWrath" is handled in Exorcism, Holy Shock, Consecration (and Holy Wrath once implemented)
+
+	paladin.registerHammerOfTheRighteous()
+	// "RuneWristImprovedHammerOfWrath" is handled Hammer of Wrath
+	// "RuneWristPurifyingPower" is handled in Exorcism
+}
+
+func (paladin *Paladin) fanaticism() float64 {
+	return core.TernaryFloat64(paladin.HasRune(proto.PaladinRune_RuneHeadFanaticism), 18, 0) * core.CritRatingPerCritChance
```
kayla-glick, line 29: 🚫 I have the same issue about the method name here. It's not clear from usage that this is providing a crit value.

If you want too you can add a shared function for grabbing paladin holy spell bonus crit

```
func (paladin *Paladin) bonusHolySpellCritRating() {
  return paladin.holyPowerCritRating() + paladin.fanaticismCritRating()
}
```

sim/paladin/hammer_of_wrath.go
```diff
@@ -0,0 +1,75 @@
...
+			},
+
+			ApplyEffects: func(sim *core.Simulation, target *core.Unit, spell *core.Spell) {
+				damage := sim.Roll(rank.damageLow, rank.damageHigh) + 0.429*spell.SpellDamage()
+				spell.CalcAndDealDamage(sim, target, damage, spell.OutcomeRangedHitAndCrit)
+
+				// should be based on target.CurrentHealthPercent(), which is not available
+				if hasImprovedHammerOfWrath && sim.CurrentTime >= time.Duration(0.9*float64(sim.Duration)) {
+					cd.Reset()
+				}
```
kayla-glick, line 71: ❓ Why is `target.CurrentHealthPercent()` not available here? My linter isn't complaining when I try to use this in a spell's `ApplyEffects`

image

sim/paladin/holy_wrath.go
```diff
@@ -0,0 +1,90 @@
...
+					if hasPurifyingPower || (target.MobType == proto.MobType_MobTypeDemon || target.MobType == proto.MobType_MobTypeUndead) {
+						damage := sim.Roll(rank.damageLow, rank.damageHigh) + 0.19*spell.SpellDamage()
+						result := spell.CalcDamage(sim, target, damage, spell.OutcomeMagicHitAndCrit)
+						results = append(results, result)
+					}
+				}
+
+				for _, result := range results {
+					spell.DealDamage(sim, result)
+				}
```
kayla-glick, line 84: ❓ What's the reason for separating out calculating and dealing the results here?

sim/paladin/paladin.go
```diff
@@ -48,8 +48,8 @@ type Paladin struct {
 	Judgement         *core.Spell
 	DivineFavor       *core.Spell
 	// HolyShield            *core.Spell
-	// HammerOfWrath         *core.Spell
-	// HolyWrath             *core.Spell
+	// HammerOfWrath         []*core.Spell
+	// HolyWrath             []*core.Spell
```
kayla-glick, line 52: Should these be un-commented and assigned in their respective spell initializers?

kayla-glick (changes requested):
Thanks for adding these 🙏 there's a few changes I'd like to see changed but then it's good to go

sim/paladin/hammer_of_wrath.go
rosenrusinov: For target to have health percent it means you need to run a health based sim. Most users run time based sim so the time*0.9 is the equivelent for it

kayla-glick: Oh weird TIL

sim/paladin/runes.go
vigo2: Same as above, but can change.

proto/paladin.proto
```diff
@@ -74,6 +74,13 @@ enum PaladinRune {
 	RuneFeetGuardedByTheLight = 415059;
 	RuneFeetSacredShield = 412019;
 	RuneFeetTheArtOfWar = 426157;
+
+	RuneHeadFanaticism = 429142;
+	RuneHeadWrath = 429139;
+
+	RuneWristHammerOfTheRighteous = 407632;
+	RuneWristImprovedHammerOfWrath = 429152;
+	RuneWristPurifyingPower = 429144;
```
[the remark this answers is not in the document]
vigo2: Ok

sim/paladin/holy_wrath.go
vigo2: That's the common pattern for multi-target spells, so procs don't interfere with damage calculation.

sim/paladin/paladin.go
vigo2: I think this pattern is outdated: these are assigned, but never used, since rotation isn't handled in code anymore.

There's even a pre-existing comment to this effect: "Active abilities and shared cooldowns that need externally manipulated". But almost none of these are "externally manipulated", luckily - since cross file references aren't making the code any easier to read.

I think these should be removed for all classes, unless such references are actually needed. And they should _definitely_ not be public.

sim/paladin/talents.go
vigo2: This method is only used to set "BonusCritRating", so I think the longer name is rather redundant. But I'll change to house style.

kayla-glick: Thanks 🙏 it just adds clarity for those who aren't as familiar with the class's mechanics

sim/paladin/holy_wrath.go
kayla-glick: Ooh ok. I might need to change some other multi-target spells then

sim/paladin/paladin.go
kayla-glick: I guess yeah if they're not actively being used they aren't necessarily needed. They can always be added as needed. They just store pointers so they're not a massive memory hog but it's something

kayla-glick (approved):
No blockers thanks for updating the method names

PR #592, opened 2024-04-05
Bonusdmg refactor

FelixPflaum:
Move all bonus damage handling to CalcDamage functions if possible. Reason being that bonus coefficient needs to be known anyway to properly add bonus damage taken.

This is done by defining BonusCoefficient on the Spell and the DoT struct. Bonus damage done is then added in CalcDamage, CalcPeriodicDamage and on dot Snapshot, which now uses a helper function if possible.

### Commits that change test results:
- bf4342c: Fix Warrior SCHOOL_DAMAGE coefs
- 298d36a: Fix Priest eternal Mind Spike crit bonus
- 5cd4985: Fix Warlock CoA ramp up and snapshot
- ab2e998: Shaman: Remove bonus weapon dmg where coef=0
- 93e4258: Hunter: Update bonus coefs for weapon spells
- 638b429: Results conflict because I made Rolling Thunder a spell that can crit, Affli and Enh is just previous changes again.
- 0251f09: Change default DoT rollover behavior to not snapshot anything

### Further changes
- Use DamageDoneByCasterMultiplier for Lake of Fire.
- Sweeping Strikes should be more like ingame now


sim/core/spell_result.go
```diff
@@ -292,16 +287,40 @@ func (spell *Spell) calcDamageInternal(sim *Simulation, target *Unit, baseDamage
...
+func (dot *Dot) Snapshot(target *Unit, baseDamage float64, isRollover bool) {
+	dot.SnapshotBaseDamage = baseDamage
+	if dot.BonusCoefficient > 0 {
+		dot.SnapshotBaseDamage += dot.BonusCoefficient * dot.Spell.SpellDamage()
+	}
+
+	if !isRollover {
+		attackTable := dot.Spell.Unit.AttackTables[target.UnitIndex][dot.Spell.CastType]
+		dot.SnapshotAttackerMultiplier = dot.Spell.AttackerDamageMultiplier(attackTable)
+	}
```
rosenrusinov, line 315: Rollovers in SOD dont update anything except the duration. Even SP is snapshot from the original cast

rosenrusinov (changes requested):
Lots of `OnSnapshot` funcs are now ignoring their `isRollover` parameter and just passing a hardcoded `false` to your new snapshot utility which is wrong. Your change should not have any effects on dot rollovers.

sim/core/dot.go
vigo2, line 1: I still wonder if it's easier to split those relatively few DD + DOT spells into two (with tags). Most DOTs are "pure", and most bleeds would have `BonusCoefficient: 0` here, so there's no way to warn if someone accidentally omits this (or just use the applier spell's `BonusCoefficient` as default).
Auto-combining applier and DOT effect as it's done now doesn't feel very natural to start with, so maybe DOTs effects could always have their own spells.

sim/core/spell_result.go
rosenrusinov: Yeah, that could also be done and the `OnShapshot` can instead be called only on application and not on rollover from the dot code

rosenrusinov: Tho when I think about it again blizzard can decide to make rollovers recalculate sp again and then we will have to revert it back. I think its better to leave the rollover support in the snapshot function and fix it to be properly passed to the shortcut

sim/core/spell_result.go
```diff
@@ -97,64 +93,57 @@ func (spell *Spell) PhysicalCritCheck(sim *Simulation, attackTable *AttackTable)
...
+				bonusDamage = power
 			}
 		}
-		return maxPower
 	}
+
+	return bonusDamage +
+		spell.Unit.GetStat(stats.SpellPower) +
+		spell.Unit.PseudoStats.MobTypeSpellPower +
+		spell.Unit.GetStat(stats.SpellDamage)
```
vigo2, line 146: What's the difference between stats.SpellPower and stats.SpellDamage? SpellPower for "increases damage and healing done", and SpellDamage for "increases spell damage"?
For the latter, I don't know if effects of this kind actually exist, e.g. Wizard Oils do in fact increase healing done as well.

rosenrusinov: Arcane Elixir from the top of my head but there are others as well

sim/core/spell_result.go
FelixPflaum: Changed the default rollover behavior. 

Imho DoTs could be simplified by adding default behavior for things like OnApplication, OnRollover and OnStackChange. Especially if AP would be handled similar to how SP is now handled in this PR, there would be no reason to do anything custom for most spells. The separation of Spell and DoT is also not the best choice, because both have common data and logic when cast/applied. There should really be SpellEffects, pretty much like the game does it. But that's all far beyond this PR.

sim/core/spell_result.go
```diff
@@ -97,64 +97,57 @@ func (spell *Spell) PhysicalCritCheck(sim *Simulation, attackTable *AttackTable)
...
 
-func (spell *Spell) SpellSchoolPower() float64 {
 	switch spell.SchoolIndex {
 	case stats.SchoolIndexNone:
 		return 0
 	case stats.SchoolIndexPhysical:
-		// Return correct value if ever used for a physical spell.
+		// PseudoStats.BonusDamage is physical "spell power", just return that here.
+		// TODO: Do "MobTypeSpellPower" effects for physical exist? E.g. something like "Do x extra weapon damage against y type"?
+		// If yes then it needs to be handled here and for the multi school case below.
```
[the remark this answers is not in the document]
FelixPflaum: They are on the weapon, so I'd guess they would work like the other weapon damage increases on weapons, just only against beasts. That would be handled outside of any spell code anyway.

And I'm pretty sure that aura effect is simply physical only bonus damage against creature types, if used on normal auras.

Considering the wording on the old relic https://www.wowhead.com/classic/spell=2635/beast-slayer it was probably intended to be bonus damage for all schools, but was then changed to be physical only which its usage on https://www.wowhead.com/classic/item=209678/mantle-of-the-thresher-slayer implies. For magic bonus damage the effect MOD_FLAT_SPELL_DAMAGE_VERSUS was created, e.g. for https://www.wowhead.com/classic/item=23207/mark-of-the-champion

If that unimplemented BFD shoulder remains the only item it can probably be ignored for now.

sim/core/spell.go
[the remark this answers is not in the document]
FelixPflaum: Yes, will probably open a PR that adds bonus dmg taken on top of this one. Adding that changes all test results and makes it hard to keep track of what changed what :D

sim/core/dot.go
FelixPflaum: As mentioned in that other comment the way it's done now is not optimal and a bit awkward for sure. Doing bonus damake taken right now that became even more apparent, because there's 3 different places to look for the BonusCoefficient there. Spell.BonusCoefficient, Spell.Dot(target).BonusCoefficient and Spell.aoeDot.BonusCoefficient xD

sim/core/spell_result.go
FelixPflaum: There are no effects in the game, Arcane Elixir also just adds a bonus for all magical schools. It's just used as a convenience in the sim to not have to add a bonus for multiple schools.

FelixPflaum:
If the state of DoT (rollover) handling is acceptable/correct now, then this is fit for adding bonus damage taken as a followup from my side.

sim/core/dot.go
vigo2: Just for reference: I've recently split Saber Slash into an application and a "tick" spell - there's a bit of SpellConfig duplication, but it makes hit/crit counting etc. in the "Results" tab a lot more obvious. The "Timeline" doesn't play well with this setup (yet ;)), but again: DD + DOT effects aren't all that common. It's obviously a "poor man's" handling of multiple SpellEffects.

sim/core/spell.go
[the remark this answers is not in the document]
vigo2: Replacing those two shouldn't change any tests: you're already adding `PseudoStats.BonusDamage` in the pipeline, so any remaining `BonusWeaponDamage()` usages in ApplyEffects() should be wrong (adding it "manually" _and_ in the pipeline), and the physical bonus dmg taken case is already supported.

sim/core/spell_result.go
vigo2: I'm still voting for "Snapshot()" to support only the default case, and ignore rollovers or stacking DOTs. It's a very helpful utility function as it is, and it's easier to use if it just does one thing.
Rollovers simply wouldn't call it at all, stacking DOTs could use it for the initial application.

sim/core/spell_result.go
vigo2: I think SpellPower (as in: a single SpellEffect affecting both damage and healing spells) isn't a thing in SOD: all "increases damage and healing done" spells have two effects "inside" (Mod Healing Done (X) and Mod Damage Done (X)).
Arcane Elixir and a few others indeed only have "Mod Damage Done (X)", but obviously, `stats.SpellPower` is redundant with setting both `stats.SpellDamage` and `stats.SpellHealing`. There's not many dynamic effects increasing both, so there wouldn't be many changes if `stats.SpellPower` were removed.

This is obviously not for _this_ PR, but it would help the pipeline quite a bit, if <school>Power were removed from stats, and merged with their physical counterpart (`PseudoStats.BonusDamage`). This would also make the "damage done" and "damage taken" pipeline more symmetric, going forward.

sim/core/attack.go
```diff
@@ -366,11 +366,10 @@ func (unit *Unit) EnableAutoAttacks(agent Agent, options AutoAttackOptions) {
 
 		DamageMultiplier: 1,
 		ThreatMultiplier: 1,
+		BonusCoefficient: TernaryFloat64(options.MainHand.GetSpellSchool() == SpellSchoolPhysical, 1, 0),
```
vigo2, line 369: Not for this PR, and pre-existing: SpellSchool should always be Physical, and by extension, BonusCoefficient always be 1 here.

The SpellConfig can be modified from the outside, anyhow, so Shadowfiend (which might just as well have BonusCoefficient 1, guessing from "Observed 6 times stronger AP scaling then expected") and even more so possible enemy mobs could just customize outside of default options.

sim/core/spell.go
```diff
@@ -36,17 +36,16 @@ type SpellConfig struct {
 	Cast               CastConfig
 	ExtraCastCondition CanCastCondition
 
-	BonusHitRating       float64
-	BonusCritRating      float64
-	BonusSpellPower      float64
-	BonusExpertiseRating float64
-	BonusArmorPenRating  float64
+	BonusHitRating  float64
```
vigo2, line 39: Thanks for also removing BonusExpertiseRating and BonusArmorPenRating. Stats and PseudoStats _really_ need a cleanup

sim/core/spell_result.go
```diff
@@ -76,10 +76,6 @@ func (spell *Spell) RangedAttackPower(target *Unit) float64 {
 		target.PseudoStats.BonusRangedAttackPowerTaken
 }
 
-func (spell *Spell) BonusWeaponDamage() float64 {
```
vigo2, line 79: So it is gone :) Ignore the related remark about it being not longer necessary elsewhere.

sim/core/spell_result.go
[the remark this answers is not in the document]
vigo2: Definitely.

sim/core/spell_result.go
```diff
@@ -292,16 +283,35 @@ func (spell *Spell) calcDamageInternal(sim *Simulation, target *Unit, baseDamage
...
+	}
 	return spell.calcDamageInternal(sim, target, baseDamage, attackerMultiplier, true, outcomeApplier)
 }
 func (dot *Dot) CalcSnapshotDamage(sim *Simulation, target *Unit, outcomeApplier OutcomeApplier) *SpellResult {
 	return dot.Spell.calcDamageInternal(sim, target, dot.SnapshotBaseDamage, dot.SnapshotAttackerMultiplier, true, outcomeApplier)
 }
 
+func (dot *Dot) Snapshot(target *Unit, baseDamage float64, isRollover bool) {
+	// Rollovers in SoD don't seem to update anything
+	if !isRollover {
```
vigo2, line 304: nit: invert for early out.

vigo2: And once again: I really like this helper, but the "isRollover" support should just be removed.

sim/core/spell_result.go
```diff
@@ -396,12 +406,28 @@ func (spell *Spell) calcHealingInternal(sim *Simulation, target *Unit, baseHeali
...
+	}
 	return spell.calcHealingInternal(sim, target, baseHealing, spell.CasterHealingMultiplier(), outcomeApplier)
 }
 func (dot *Dot) CalcSnapshotHealing(sim *Simulation, target *Unit, outcomeApplier OutcomeApplier) *SpellResult {
 	return dot.Spell.calcHealingInternal(sim, target, dot.SnapshotBaseDamage, dot.SnapshotAttackerMultiplier, outcomeApplier)
 }
 
+func (dot *Dot) SnapshotHeal(target *Unit, baseHealing float64, isRollover bool) {
+	// Rollovers in SoD don't seem to update anything
+	if !isRollover {
```
vigo2, line 420: nit, again: invert for early-out

vigo2 (approved):
This will bitrot fast, so approved. Two minor nits (and the "Snapshot()" wish) remain, I gather you'll add proper DamageTaken in the next, very short PR?

For this you _could_ simply replace `spell.Flags.Matches(SpellFlagIncludeTargetBonusDamage)` with `spell.BonusCoefficient > 0` in `applyTargetModifiers`, this shouldn't change any tests, unless the Flag was misplaced before. Probably worth a try ;)

rosenrusinov (approved):

rosenrusinov:
This is already bigger then it should have been and changing more stuff then its original scope. I'm merging as is. Any further changes can be done in a separate PR (hopefully not growing the size of this one)

PR #594, opened 2024-04-05
Update crit and dodge scaling from agi and int

raethkcj:
Collected in-game using RatingBuster: https://github.com/raethkcj/RatingBuster/blob/master/libs/StatLogic/Vanilla_Logic.lua#L49-L201


vigo2:
Excellent. The base dodge, melee crit, and spell crit values are just as reliable, I'd guess?

raethkcj:
I believe base crit and spell crit are already in and correct, I didn't notice it was missing base dodge. Base dodge is all equal to base melee crit in vanilla so it's pretty easy!

PR #631, opened 2024-04-09
Mage/Druid hotfixes

kayla-glick:


sim/druid/talents.go
```diff
@@ -265,7 +265,7 @@ func (druid *Druid) applyOmenOfClarity() {
 				return
 			}
 
-			if spell.Flags.Matches(SpellFlagOmen) {
+			if spell.Flags.Matches(SpellFlagOmen) && spell.DefaultCast.Cost > 0 {
```
rosenrusinov, line 268: `spell.CurCast.Cost`

kayla-glick: Oh dang you're right

kayla-glick: Actually no because omen makes the spell not cost anything but this specifically covers a hotfix where free wraths from the Fury of Stormrage elune do not consume clearcasting now

PR #649, opened 2024-04-10
Prepull Combo Point APL Addition

sanguinerarogue:
Creating ability to add combo points to the APL via Honor Among Thieves.  I had some issues with the log bugging out if sent a nil value for the SpellID so right now it is assigned to HaT.  Haven't touched the APL at all so would appreciate a review.

image

image

### DIFF
--- a/sim/core/apl_actions_misc.go
+++ b/sim/core/apl_actions_misc.go
@@ -136,7 +136,46 @@ func (action *APLActionActivateAuraWithStacks) Execute(sim *Simulation) {
 func (action *APLActionActivateAuraWithStacks) String() string {
 	return fmt.Sprintf("Activate Aura(%s) Stacks(%d)", action.aura.ActionID, action.numStacks)
 }
+//START
+type APLActionAddComboPoints struct {
+	defaultAPLActionImpl
+	character *Character
+	numPoints int32
+}
 
+func (rot *APLRotation) newActionAddComboPoints(config *proto.APLActionAddComboPoints) APLActionImpl {
+	character := rot.unit.Env.Raid.GetPlayerFromUnit(rot.unit).GetCharacter()
+	numPoints, err := strconv.Atoi(config.NumPoints)
+	
+	if err != nil {
+		numPoints = 0
+	}
+	return &APLActionAddComboPoints{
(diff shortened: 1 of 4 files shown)


sim/core/apl_actions_misc.go
```diff
@@ -137,6 +137,44 @@ func (action *APLActionActivateAuraWithStacks) String() string {
...
+	return true
+}
+
+func (action *APLActionAddComboPoints) Execute(sim *Simulation) {
+	numPoints := strconv.Itoa(int(action.numPoints))
+
+	if sim.Log != nil {
+		action.character.Log(sim, "Adding combo points (%s points)", numPoints)
+	}
+	metrics := action.character.NewComboPointMetrics(ActionID{SpellID: 432264})
```
kayla-glick, line 169: @sanguinerarogue and I talked about this before but this spell ID doesn't really make sense here both for Rogue (who may not even have the Rune equipped) and also since this will be useful for Feral. The problem is that you have to provide an ActionID, but you can't use a blank ID or `SpellID: 0` without breaking things apparently.

rosenrusinov: @sanguinerarogue What you should do is create a new dummy id like we do for Attack/Move/Wait/Etc. It is added in common.proto in `enum OtherAction` and then add it to the switch case in `action_id.ts`

sanguinerarogue:
@rosenrusinov thank you for that lead.  I've gone ahead and done so to create the add combo points ID.  

image

However, now the resources tab is generating a line for each simulation for this addition (seen below).  Let me know if you have seen this before or can point me in the right direction. 

image

rosenrusinov:
@sanguinerarogue 
> However, now the resources tab is generating a line for each simulation for this addition (seen below). Let me know if you have seen this before or can point me in the right direction.

That is because you are creating a new metric on every Execute. The metric should be created once (in the apl action creation or even in the energy bar whatever feels better) and stored in the action struct and used in the execute

sanguinerarogue:
@rosenrusinov Thank you for the help! It looks resolved now if you don't mind taking another look.

kayla-glick (approved):

PR #652, opened 2024-04-10
Implemented TNT rune for hunters

ncberman:
Adding the TNT rune for hunters

[Link to Rune](https://www.wowhead.com/classic/spell=428717/t-n-t)

Implementation Details:
- No code changes in runes.go
- Each affected spell has had a damageMult variable which defaults to 1.0 and is multiplied by 1.1 when the rune is equipped
- Explosive Shot has damageMult applied wherever base damage is calculated
- Immolation trap has damageMult applied on DoT snapshot
- Explosive trap has damageMult applied on initial damage as well as on the DoT snapshot


sim/hunter/explosive_shot.go
```diff
@@ -62,7 +67,7 @@ func (hunter *Hunter) registerExplosiveShotSpell(timer *core.Timer) {
 			NumberOfTicks: 2,
 			TickLength:    time.Second * 1,
 			OnSnapshot: func(sim *core.Simulation, target *core.Unit, dot *core.Dot, isRollover bool) {
-				baseDamage := sim.Roll(baseLowDamage, baseHighDamage) + 0.039*dot.Spell.RangedAttackPower(target)
+				baseDamage := ( sim.Roll(baseLowDamage, baseHighDamage) + 0.039*dot.Spell.RangedAttackPower(target) ) * damageMult
```
kayla-glick, line 70: I think all of these multipliers in these spell files should be applied to the `DamageMultiplier` field since they're not base damage multipliers as far as I can tell. That's effectively what you're doing by multiplying each damage calculation like this

kayla-glick (changes requested):

sim/hunter/explosive_shot.go
ncberman: Didn't realize we had that field available, oops! I'll apply the bonus there instead

sim/hunter/explosive_shot.go
```diff
@@ -95,3 +95,11 @@ func (hunter *Hunter) registerExplosiveShotSpell(timer *core.Timer) {
...
 	})
 }
+
+func (hunter *Hunter) calcExplosiveShotDamageMultipliers() float64 {
+	damageMultiplier := 1.0
+	if hunter.HasRune(proto.HunterRune_RuneBracersTNT) {
+		damageMultiplier *= 1.1
+	}
+	return damageMultiplier
+}
```
rosenrusinov, line 105: You can remove this as its not used anywhere

kayla-glick (approved):

Issue #413, opened 2024-03-17
Crash Report 1253835652


Tonnepon:
Link:
https://wowsims.github.io/sod/hunter/#eJztVE9IFFEYn2/+7Lz9dkefL8u3r7B1QZoGXWbXFtTLTh5CIkIi0GuSkmEYCUHHFhI7JZ3UQyxEWIoQHqIoMKRQO9nBKOogQZfwsNAhumhv/uxSpw4Zdeh7zPe++X7f+73fN29m0CIqVV3oBA+mVOixOGwBPITzF2ANgCliOY6xvtFz14auUJ3badUpAep0Z8qSfqvk+8fSm/RZyeKvY4j0ST//GLO3d7lMTswA/wqYoKsSnTTtdzscCX1Vsuzyrh9tymhuMSUL3gfL7ZcyrdG5YUn7dNrf6NuXpPTL9wEVOUy6LeumjXySJmyjU5sE7TlA94SOW6ry3/6osTL8bQm/ss+q0mNS4DAJcLK+PVeQw+1wCzn/6lMHyKBycVXjGjZiXJhokMqYrWdkgFp7YRwPRNmNMVvNxFGG7W5W5psQBcEYuXcXbC0C8hJgbcLBNGvGBqeexJg6qIiAq6vQivEliKFOlhdTZ6KZ6ULtVlhKNKGVSVQLPr0Vwp8X7ggmBA+hfKBtxDZENWBSQVAnvyCWFs3IHEqAGSuqSgxhlHR/qmlkR0QrCodLUfF5iJaJQHhHvnBpvNYl6xUnsJt1Yspp+rlYcqGey3aM40EnJaHEPNToQ9CVYLDlo10ut/RFNTKG1KkjevAsZA2qhVZR7T8r2rCFHf6xe5nGukwScQmqx1E7F9YsDkVtqj1KJMh1Q74PZbi8BjejNzLrzYKyAvvZPqJRg0NGORoClWKvmVnR4uHdaa8hDBq91OyMb+vFrjCzUWwZDuxNMX9bIw8WLEZODV0dGk0fc7mTNp1/7P9Sue7tCQ+Wf4vnRbLh+J7okJYLvHfL6xoZq9xYP7tZrFJ7A/Ad4Z6guw==

RNG Seed: 980551388

runtime error: invalid memory address or nil pointer dereference
Stack Trace:
goroutine 11 [running]:
runtime/debug.Stack()
[...]

PR #654, opened 2024-04-11
Fix Attack Table when hunter has no main-hand equipped

kayla-glick:
Resolves https://github.com/wowsims/sod/issues/413

### DIFF
--- a/sim/core/environment.go
+++ b/sim/core/environment.go
@@ -226,7 +226,11 @@ func (env *Environment) setupAttackTables() {
 				character := env.Raid.GetPlayerFromUnit(attacker).GetCharacter()
 				weapons := []*Item{character.GetMHWeapon(), character.GetOHWeapon(), character.GetRangedWeapon()}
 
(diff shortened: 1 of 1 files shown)


sim/core/environment.go
```diff
@@ -226,7 +226,11 @@ func (env *Environment) setupAttackTables() {
 				character := env.Raid.GetPlayerFromUnit(attacker).GetCharacter()
 				weapons := []*Item{character.GetMHWeapon(), character.GetOHWeapon(), character.GetRangedWeapon()}
 
-				if character.Class == proto.Class_ClassMage || character.Class == proto.Class_ClassPriest || character.Class == proto.Class_ClassWarlock || weapons[0] == nil {
+				if weapons[0] == nil {
+					attacker.AttackTables[idx][proto.CastType_CastTypeMainHand] = NewAttackTable(attacker, defender, weapons[0])
+				}
```
rosenrusinov, line 231: I think this if can be removed and in the else part of the next if you can just remove the `if weapon != nil` check and create attack tables even for nil weapons. Looking at the `NewAttackTable` function it will work even with a nil weapon as the only place the weapon is used is for getting the weapon skill and that part has its own nil check for the weapon that returns 0 if its nil

rosenrusinov:
Fixes #413

PR #672, opened 2024-04-14
Update idol of the dream

kayla-glick:
Slight alternative approach to https://github.com/wowsims/sod/pull/645 but I'd like to retain the other QoL docs improvements


sim/druid/shred.go
```diff
@@ -88,20 +80,7 @@ func (druid *Druid) registerShredSpell() {
...
-
-			/*
-				if druid.AssumeBleedActive || druid.Rip.Dot(target).IsActive() || druid.Rake.Dot(target).IsActive() || druid.Lacerate.Dot(target).IsActive() {
-					modifier *= 1.0 + (0.04 * float64(druid.Talents.RendAndTear))
-				}
-			*/
-
-			spell.DamageMultiplier *= modifier
 			baseres := spell.CalcDamage(sim, target, baseDamage, spell.OutcomeExpectedMagicAlwaysHit)
-			spell.DamageMultiplier /= modifier
```
FelixPflaum, line 104: Need to restore mangle dmg increase for expected dmg too.

sim/druid/items.go
```diff
@@ -31,6 +32,21 @@ func init() {
...
+			if spell.SpellCode == SpellCode_DruidShred {
+				spell.DamageMultiplier += 2 * .02
+			}
+
+			// TODO: Also boosts the damage of swipe by 2%
+			// if spell.SpellCode == SpellCode_DruidSwipe {
+			// 	spell.DamageMultiplier *= 1.02
+			// }
+		})
+	})
```
FelixPflaum, line 48: After some testing the idol seems to work the following for Shred:
1. The added flat damage is increased by 2%.
2. The base multiplier (225%) is increased by 2%, then the additive 75% from SoD are then added on top.
Because of that it needs to happen in Shred code after all.

rosenrusinov:
Done in #679

PR #697, opened 2024-04-18
Fix warlock pet mp5 inheritance

nathro:
So, thorough in game testing has revealed that rather than inheriting a percentage of your personal mp5, the pets are actually scaling mp5 as a percentage of your intellect. This updates the inheritance to the correct value/stat.


rosenrusinov (approved):

rosenrusinov:
@nathro need to update with latest master and update tests again as you had changes to the same stuff in different PRs

PR #724, opened 2024-04-23
Hunter Pet - Perf fix

vigo2:
[druid] fixed Gale Wind, and added a comment explaining the implementation irregularity

[hunter] fix pet rotation using Cast() w/o checking CanCast(), since that's a huge performance trap

### DIFF
--- a/sim/hunter/pet.go
+++ b/sim/hunter/pet.go
@@ -1,6 +1,7 @@
 package hunter
 
 import (
+	"fmt"
 	"time"
 
 	"github.com/wowsims/sod/sim/core"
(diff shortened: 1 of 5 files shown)


sim/hunter/pet.go
```diff
@@ -215,32 +216,43 @@ func (hp *HunterPet) ExecuteCustomRotation(sim *core.Simulation) {
...
+	// using Cast() directly is very expensive, since cast failures are logged, involving string operations
+	tryCast := func(spell *core.Spell) bool {
+		if !spell.CanCast(sim, target) {
+			return false
+		}
+		if !spell.Cast(sim, target) {
+			panic(fmt.Sprintf("Cast failed after CanCast() for spell %d", spell.SpellID))
+		}
+		return true
+	}
```
kayla-glick, line 228: Do you have numbers on the performance comparison here? This changes all 3 hunter spec results - albeit by very small amounts - so I just want to make sure everything checks out

PR #727, opened 2024-04-23
Paladin Updates

vigo2:
[paladin] allow SoR to crit
[paladin] Crusader Strike now refreshes all judgements on the target to 30s 
[paladin] applying SoM keeps the previous seal lingering for 400ms, so the next auto might trigger both seals at once

The CS change is more or less ineffective outside of raid simming, and could be guarded accordingly.

Does twisting allow to apply two Judgements at once? (since both Seal auras are active at once, both may react to the generic Judgement cast, and trigger their specific Judgements).


sim/paladin/sor.go
```diff
@@ -147,7 +150,7 @@ func (paladin *Paladin) registerSealOfRighteousness() {
 			},
 
 			ApplyEffects: func(sim *core.Simulation, _ *core.Unit, _ *core.Spell) {
-				paladin.ApplySeal(aura, judgeSpell, sim)
+				paladin.applySeal(aura, judgeSpell, sim)
```
hypernormalisation, line 153: As per discord convo, applyTwistableSeal

sim/paladin/sotc.go
```diff
@@ -98,7 +98,7 @@ func (paladin *Paladin) registerSealOfTheCrusader() {
 			},
 
 			ApplyEffects: func(sim *core.Simulation, _ *core.Unit, _ *core.Spell) {
-				paladin.ApplySeal(aura, judgeSpell, sim)
+				paladin.applySeal(aura, judgeSpell, sim)
```
hypernormalisation, line 101: And here

sim/paladin/soc.go
```diff
@@ -145,7 +145,7 @@ func (paladin *Paladin) registerSealOfCommand() {
 			},
 
 			ApplyEffects: func(sim *core.Simulation, _ *core.Unit, _ *core.Spell) {
-				paladin.ApplySeal(aura, judgeSpell, sim)
+				paladin.applySeal(aura, judgeSpell, sim)
```
hypernormalisation, line 148: And here

sim/paladin/soc.go
hypernormalisation, line 1: We'll need a way for SoC to proc SoM but no other seals following the change in-game late last night. Probably a custom flag being added to the SoC proc spell would suffice, that the SoM proc spell can watch for in its proc mask? If you can think of a more elegant way go with whatever you like.

hypernormalisation (changes requested):

hypernormalisation (approved):

PR #734, opened 2024-04-25
Rogue stuff

vigo2:
[rogue] allow EA casts over existing "MajorArmorReduction" auras, unless they're stronger
[rogue] Saber Slash and Sinister Strike now deal an extra 20% damage per stack of Saber Slash applied
[various] slight refactoring for spells with dynamically changing damage modifiers

### DIFF
--- a/sim/warlock/immolate.go
+++ b/sim/warlock/immolate.go
@@ -87,10 +87,10 @@ func (warlock *Warlock) getImmolateConfig(rank int) core.SpellConfig {
 		},
 
 		ApplyEffects: func(sim *core.Simulation, target *core.Unit, spell *core.Spell) {
-			imprImmoMult := 1 + 0.05*float64(warlock.Talents.ImprovedImmolate)
-			spell.DamageMultiplier *= imprImmoMult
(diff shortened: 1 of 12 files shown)


sim/warlock/immolate.go
```diff
@@ -87,10 +87,10 @@ func (warlock *Warlock) getImmolateConfig(rank int) core.SpellConfig {
 		},
 
 		ApplyEffects: func(sim *core.Simulation, target *core.Unit, spell *core.Spell) {
-			imprImmoMult := 1 + 0.05*float64(warlock.Talents.ImprovedImmolate)
-			spell.DamageMultiplier *= imprImmoMult
+			oldMultiplier := spell.DamageMultiplier
+			spell.DamageMultiplier *= 1 + 0.05*float64(warlock.Talents.ImprovedImmolate) // TODO should most likely just be done statically (?)
```
rosenrusinov, line 91: Should not apply to the Dot part of the spell. which is why its done like this. If the dot spell is taken out as a different one then we can apply statically to the initial spell

PR #751, opened 2024-05-04
Adds weapon specialization pickers @ 60

emrus-main:
Made a pickable weapon specialization which is exclusive with racials for next phase.


kayla-glick:
These ended up being implemented as ring runes so I'm going to close this out

PR #759, opened 2024-05-11
Missing warrior talents

emrus-main:
Adds Enrage and Defiance talents
Fixes Imp Cleave typo
Reverts Shield Spec, and Imp Slam to classic version
Adds stance & shield requirements to Devastate and Shield Wall
Implemented vanilla version of Shield Block, and added Imp Shield Block.

Now only missing: Blood Craze, Piercing howl and the talents that do not work vs bosses.

Note: Shield block is now available for all warriors. But, it is disabled with regard to auto cooldown usage.

### DIFF
--- a/sim/warrior/shield_block.go
+++ b/sim/warrior/shield_block.go
@@ -7,22 +7,27 @@ import (
 	"github.com/wowsims/sod/sim/core/stats"
 )
 
-// TODO: Classic Update
 func (warrior *Warrior) RegisterShieldBlockCD() {
 	actionID := core.ActionID{SpellID: 2565}
-	cooldownDur := time.Second * 60
+	cooldownDur := time.Second * 5
 
 	warrior.ShieldBlockAura = warrior.RegisterAura(core.Aura{
-		Label:    "Shield Block",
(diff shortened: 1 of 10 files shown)


sim/warrior/shield_block.go
```diff
@@ -7,29 +7,37 @@ import (
...
+	cooldownDur := time.Second * 5
 
 	warrior.ShieldBlockAura = warrior.RegisterAura(core.Aura{
-		Label:    "Shield Block",
-		ActionID: actionID,
-		Duration: time.Second * 10,
+		Label:     "Shield Block",
+		ActionID:  actionID,
+		Duration:  time.Second * time.Duration(5+[]float64{0, 0.5, 1, 2}[warrior.Talents.ImprovedShieldBlock]),
+		MaxStacks: 1 + []int32{0, 1, 1, 1}[warrior.Talents.ImprovedShieldBlock],
```
kayla-glick, line 18: This could just be a `core.Ternary` and either 1 or 2 depending on if the talent is > 0

kayla-glick (approved):

Issue #798, opened 2024-06-13
[Phase 4] Set Bonuses


kayla-glick:
Tracking set bonuses including raid tier, dungeon tier, pvp gear, etc.

# General
Unsure if any of these will receive a SoD update

### Crafted
- [x] https://www.wowhead.com/classic/item-set=143/devilsaur-armor
- [x] https://www.wowhead.com/classic/item-set=490/green-dragon-mail
- [x] https://www.wowhead.com/classic/item-set=491/blue-dragon-mail
- [x] https://www.wowhead.com/classic/item-set=489/black-dragon-mail
- [x] https://www.wowhead.com/classic/item-set=144/ironfeather-armor
- [x] https://www.wowhead.com/classic/item-set=442/blood-tiger-harness
- [x] https://www.wowhead.com/classic/item-set=443/bloodsoul-embrace
- [x] https://www.wowhead.com/classic/item-set=444/the-darksoul
- [x] https://www.wowhead.com/classic/item-set=421/bloodvine-garb

### Drops
- [x] https://www.wowhead.com/classic/item-set=1786/deathbone-guardian
- [x] https://www.wowhead.com/classic/item-set=121/cadaverous-garb (original)
- [x] https://www.wowhead.com/classic/item-set=1782/the-postmaster
- [x] https://www.wowhead.com/classic/item-set=1787/necropile-raiment
- [x] https://www.wowhead.com/classic/item-set=1788/bloodmail-regalia
- [x] https://www.wowhead.com/classic/item-set=1785/ironweave-battlesuit
- [x] https://www.wowhead.com/classic/item-set=1783/spiders-kiss
- [x] https://www.wowhead.com/classic/item-set=1784/dalrends-arms
- [ ] https://www.wowhead.com/classic/item-set=1779/core-hounds-call
  - Effects implemented but need pet data
- [ ] https://www.wowhead.com/classic/item-set=1780/shard-of-the-gods
  - Effects implemented but need information on the servants
- [ ] https://www.wowhead.com/classic/item=228350/eskhandars-right-claw
  - Effects implemented but need pet data

### Unsure
These might be from the Naxxramas event and not relevant yet
- [ ] https://www.wowhead.com/classic/item-set=535/garb-of-the-undead-slayer
- [ ] https://www.wowhead.com/classic/item-set=534/undead-slayers-armor
- [ ] https://www.wowhead.com/classic/item-set=533/battlegear-of-undead-slaying
[...]

PR #817, opened 2024-06-25
fix: dreadmist set bonus should grant health instead of mana

moonkxn:
Probably just a typo, [proc](https://www.wowhead.com/classic/spell=450583/dark-reward) tooltip and effect are both `heal`


kayla-glick (approved):

kayla-glick:
https://github.com/wowsims/sod/issues/798

PR #821, opened 2024-06-25
Add priest tier 1 sets

kayla-glick:


kayla-glick:
https://github.com/wowsims/sod/issues/798

PR #824, opened 2024-06-28
New mage changes

kayla-glick:
- Brain Freeze proc chance raised to 20%
- Fire Blast cooldown reduced to 15s when using Overheat
- Balefire bolt is now Arcane + Fire + Frost schools
- Updated APLs accordingly with findings


kayla-glick:
https://github.com/wowsims/sod/issues/793

PR #829, opened 2024-06-28
Hunter trapfixes

ncberman:
Fixed an issue where a target could take Immolation Trap DoT damage and Explosive Trap DoT damage at the same time.

Updated hunter traps to be available all the time rather than just when trap launcher rune is equipped.
trap launcher rune now increases trap range requirements from 5 -> 35
all traps now share a cooldown unless the trap launcher rune is equipped then fire traps and frost traps have independent cooldowns.
Screenshot 2024-06-28 145823

Updated Lone Wolf rune damage bonus from 30% to 40%.
Screenshot 2024-06-28 145926

Corresponding Blue Post: https://us.forums.blizzard.com/en/wow/t/wow-classic-season-of-discovery-phase-4-ptr-development-notes-updated-6282024/1883827

### DIFF
--- a/sim/hunter/explosive_trap.go
+++ b/sim/hunter/explosive_trap.go
@@ -35,12 +35,15 @@ func (hunter *Hunter) getExplosiveTrapConfig(rank int, timer *core.Timer) core.S
 		Cast: core.CastConfig{
 			CD: core.Cooldown{
 				Timer:    timer,
-				Duration: time.Second * time.Duration(15 * hunter.resourcefulnessCooldownModifier()),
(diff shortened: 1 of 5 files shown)


ncberman:
I've verified all the changes work how I'd expect them to on my end, if the code looks good it should be ready to merge! 
I did have one hiccup when implementing the mana cost modifier to immo trap and frost trap spells being that I tried to use the Multiplier field but since the mana cost multiplier is 0.0 (100% cost reduction) the multiplier would default to 1 so I ended up baking the calculation into the flat cost.

kayla-glick (approved):

PR #837, opened 2024-06-30
Updates several rogue runes to latest patch

sanguinerarogue:
1. Shuriken Toss - Adjusted AP to 50%, added CP for targets and 30s cooldown
2. Poisoned Knife - Added energy return scaling on DP stacks, removed minimum range
3. Saber Slash - Dropped from 5 to 3 stacks, increased scaling to 33% and did 5% AP scaling per wowhead
4. Slaughter from Shadows - 60% damage increase + reduction in cost
5. Unfair Advantage - I tried several syntax to test dodge + parry but couldnt get it right, works with just dodge right now
6. Blade Dance - AP scaling not working, not sure issue

@catszeid If you could take a look at this and let me know on 5 and 6.

### DIFF
--- a/sim/rogue/backstab.go
+++ b/sim/rogue/backstab.go
@@ -32,7 +32,7 @@ func (rogue *Rogue) registerBackstabSpell() {
 		Flags:       rogue.builderFlags(),
 
 		EnergyCost: core.EnergyCostOptions{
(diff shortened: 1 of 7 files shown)


sanguinerarogue:
Fixed the issues in 5 and 6

PR #838, opened 2024-06-30
Filter items reworked in SoD + parse SoD phase information from wowhead gearplanner

kayla-glick:


ui/core/player.ts
```diff
@@ -1144,16 +1144,19 @@ export class Player<SpecType extends Spec> {
...
+				if (typeof zoneId == 'number' && zoneId != 0) {
+					zoneIds.push(zoneId);
 				}
 			}
+
+			itemData = filterItems(
+				itemData,
+				item => !item.sources.some(itemSrc => itemSrc.source.oneofKind == 'drop' && zoneIds.includes(itemSrc.source.drop.zoneId)),
+			);
 		}
```
kayla-glick, line 1160: This should be significantly more performant

PR #839, opened 2024-06-30
Hunter talent updates + Rapid Killing & Cobra Strikes rune update

ncberman:
Updated hunter MM talents according to the following blue post.
Removed crit chance facade from lethal shots (was causing a visual bug where each stat was adding 2% phys crit rather than 1%)
Screenshot 2024-06-30 112700
Also included rapid fire rune update
Screenshot 2024-06-30 114756

### DIFF
--- a/sim/hunter/flanking_strike.go
+++ b/sim/hunter/flanking_strike.go
@@ -13,6 +13,8 @@ func (hunter *Hunter) registerFlankingStrikeSpell() {
 	}
 
 	hasCatlikeReflexes := hunter.HasRune(proto.HunterRune_RuneHelmCatlikeReflexes)
+	hasCobraStrikes := hunter.pet != nil && hunter.HasRune(proto.HunterRune_RuneChestCobraStrikes)
+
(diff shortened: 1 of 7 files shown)


ncberman:
Decided to add the cobra strikes rune change as well as it seemed minor. 
Only strike ability in the sim right now is raptor strike.
Screenshot 2024-06-30 120735

sim/hunter/flanking_strike.go
```diff
@@ -65,6 +67,8 @@ func (hunter *Hunter) registerFlankingStrikeSpell() {
 		})
 	}
 
+	manaCostMultiplier := 1 - 0.02*float64(hunter.Talents.Efficiency)
```
kayla-glick, line 70: One way to handle this without updating every strike/sting etc would be to use an `OnSpellRegistered` callback and applying it to all relevant spells there based on a flag or specific spell codes. That might be a good refactor opportunity for later

kayla-glick (approved):

PR #841, opened 2024-06-30
Hunter Wyvern Strike Implementation

ncberman:
Implemented Wyvern Strike according to PTR as well as this log:
https://vanilla.warcraftlogs.com/reports/hJXxcZPvRdr64HwF#fight=3&type=damage-done&source=146&view=events

Link to spell: https://www.wowhead.com/classic/spell=458482/wyvern-strike

Also added spell ranks to some hunter spells that lacked them in the db.json to clean up the apl spell list


sim/hunter/wyvern_strike.go
```diff
@@ -0,0 +1,96 @@
...
+				tickDamage := dotDamage / float64(dot.NumberOfTicks)
+				dot.Snapshot(target, tickDamage, isRollover)
+			},
+			OnTick: func(sim *core.Simulation, target *core.Unit, dot *core.Dot) {
+				dot.CalcAndDealPeriodicSnapshotDamage(sim, target, dot.OutcomeTickCounted)
+			},
+		},
+
+		ApplyEffects: func(sim *core.Simulation, target *core.Unit, spell *core.Spell) {
+			weaponDamage := spell.Unit.MHWeaponDamage(sim, spell.MeleeAttackPower())
```
kayla-glick, line 68: According to the wowhead entry it looks like this should use normalized weapon damage
image

sim/hunter/wyvern_strike.go
```diff
@@ -0,0 +1,96 @@
...
+		Dot: core.DotConfig{
+			Aura: core.Aura{
+				Label: "WyvernStrike" + hunter.Label + strconv.Itoa(rank),
+				Tag:   "WyvernStrike",
+			},
+			NumberOfTicks: 8,
+			TickLength:    time.Second * 1,
+
+			OnSnapshot: func(sim *core.Simulation, target *core.Unit, dot *core.Dot, isRollover bool) {
+				dotDamage := dotDamageCoeff / 100 * 8 * hunter.GetStat(stats.AttackPower)
```
kayla-glick, line 58: For the `8 * hunter.GetStat(stats.AttackPower)` bit you can use the `BonusCoefficient` field inside of the `DotConfig` 

```
NumberOfTicks: 8,
TickLength: time.Second * 1,
BonusCoefficient: 8
```

kayla-glick (changes requested):
Left a few comments

ncberman:
> Left a few comments

Changed Wyvern Strike to be normalized and we talked about the bonus coeff, summary is that bonus coeff only applies bonus damage modifiers for physical spells doesn't currently work to use attack power like spell damage

sim/hunter/wyvern_strike.go
```diff
@@ -0,0 +1,94 @@
...
+	"strconv"
+	"time"
+
+	"github.com/wowsims/sod/sim/core"
+	"github.com/wowsims/sod/sim/core/proto"
+)
+
+func (hunter *Hunter) getWyvernStrikeConfig(rank int) core.SpellConfig {
+	spellId := [4]int32{0, 458436, 458481, 458482}[rank]
+	baseDamage := [4]float64{0, 3, 4, 6}[rank]
```
kayla-glick, line 13: What do you think about adding the `/100 * 8` up here instead of in the `OnSnapshot` so that the sim doesn't have to do that calculation on each snapshot?

ncberman: Yeah I think that makes sense I'll move the calculation

kayla-glick: Oh yeah I guess that's true. You could just do `dotAPCoef` to be short if you want

kayla-glick (approved):

Issue #791, opened 2024-06-13
[Phase 4] New Runes: Hunter


kayla-glick:
# TODO
- [ ] Giantstalker Pursuit set bonus https://www.wowhead.com/classic/item-set=1702/giantstalker-pursuit
- [ ] Giantstalker Prowess set bonus https://www.wowhead.com/classic/item-set=1703/giantstalker-prowess

ncberman:
#829

adocoD:
List of P4 hunter changes:

Runes
	New runes:
		- Resourcefulness = Trap mana cost -100% cd from 15 to 9s
		- Cobra Slayer = 5% chance per auto for MB proc, +5% for each auto. 100% on dodge. +40% AP on MB
		- Kill Shot = 100% weap dam + 138.435. 15s cd. Insta reset on 20% or lower
		- Wyvern Strike = 100% weap damage, 8s cd, Bleed for (3 / 100 * 8 * Attack power) over 8s

	Deleted runes:
		- Heart = baseline +200 mAP, Kill command, Invigoration

	Old rune changes:
		- TNT = Explo initial increase 50% of mAP. Immo dot increase 50% of mAP
		- Catlike Reflexes = Kill command gone
		- Rapid Killing = Also 40% Melee AS (CD -80%)
		- Check Melee Specialist
		- Check Steady Shot (100%)
		- Check DW Spec
		- Check Chimera
		- Check Explosive
		- Flanking strike = 5% all damage increase 10s, 3 stack max, pet auto 33% reset chance
		- Carve = 65% + 50% on primary (so 97,5 total)
		- Check Beast Mastery
		- Lone Wolf = 40%
		- Cobra Strike = now also from all Strikes and Mongoose Bite

Talents
	MM:
		- Efficiency = Also for Raptor, Flanking, Wyvern
		- Lethal Shots = also for all melee, possibly all crit though
		- Mortal Shots = also for MB, Raptor, Wing clip, Flanking, Wyvern
		- Trueshot = only rAP (200 rAP)

Gear
		- Tier 0, 0.5, 1 sets


Other Notes:
		- Faster OH will be better for Dual wield cause of Mongoose procs (no daggers cause of normalization)

ncberman:
Created a new PR for some of the easier changes #834

ncberman:
Have a new PR to address the minor rune changes, Rapid Fire and Cobra Strikes as well as update efficiency, lethal shots, and mortal shots talents. #839

[...]

PR #844, opened 2024-07-02
Hunter Mongoose Bite Spell / Rune + Mortal Shots talent fix

ncberman:
Previously Mortal Shots was implemented to affect ALL melee crit damage, however upon further review the talent should only affect some of the melee ABILITIES. Link: https://www.wowhead.com/classic/spell=19490/mortal-shots

Added Mongoose Bite to hunter's spellbook. Since the ability can only be activated after dodging an enemies attack the ability is only available when you have the Cobra Slayer rune which allows you to use Mongoose Bite when an enemy dodges your attack or if you proc the trigger. (Inspiration taken by warrior overpower implementation) Mongoose Bite is also unaffected by bonus physical damage such as Glowing Neuro-link Cowl.
https://www.wowhead.com/classic/spell=458393/cobra-slayer
Proc aura:
https://www.wowhead.com/classic/spell=5302/defensive-state

#791


sim/hunter/mongoose_bite.go
```diff
@@ -0,0 +1,110 @@
...
+		Label:    "Defensive State",
+		ActionID: core.ActionID{SpellID: 5302},
+		Duration: time.Second * 5,
+	})
+}
+
+func (hunter *Hunter) registerMongooseBiteSpell() {
+	if !hunter.HasRune(proto.HunterRune_RuneChestCobraSlayer) {
+		return
+	}
```
kayla-glick, line 100: Since Mongoose bite exists as a separate ability from the Cobra Slayer rune, could we separate the creation of the cobra slayer aura from registering mongoose bite? It could get in `hunter/runes.go` instead

sim/hunter/mongoose_bite.go
```diff
@@ -0,0 +1,110 @@
...
+func (hunter *Hunter) registerCobraSlayerAura() {
+	hunter.RegisterAura(core.Aura{
+		Label:    "Cobra Slayer Trigger",
+		Duration: core.NeverExpires,
+		MaxStacks: 20,
+		OnReset: func(aura *core.Aura, sim *core.Simulation) {
+			aura.Activate(sim)
+		},
+		OnSpellHitDealt: func(aura *core.Aura, sim *core.Simulation, spell *core.Spell, result *core.SpellResult) {
+			if result.Outcome.Matches(core.OutcomeDodge) {
```
kayla-glick, line 74: Shouldn't this be dodge, block, or parry?
image

sim/hunter/mongoose_bite.go
```diff
@@ -0,0 +1,110 @@
...
+
+		BonusCritRating:  float64(hunter.Talents.SavageStrikes) * 10 * core.CritRatingPerCritChance,
+		CritDamageBonus: hunter.mortalShots(),
+		DamageMultiplier: 1,
+
+		ApplyEffects: func(sim *core.Simulation, target *core.Unit, spell *core.Spell) {
+			hunter.DefensiveState.Deactivate(sim)
+
+			multiplier := 1.0
+			if stacks := hunter.RaptorFuryAura.GetStacks(); stacks > 0 {
```
kayla-glick, line 53: This probably also needs to check that the Raptor Fury rune is equipped other it will error out

sim/hunter/mongoose_bite.go
```diff
@@ -0,0 +1,110 @@
...
+			},
+		},
+
+		ExtraCastCondition: func(sim *core.Simulation, target *core.Unit) bool {
+			return hunter.DistanceFromTarget <= 5 && hunter.DefensiveState.IsActive()
+		},
+
+		BonusCritRating:  float64(hunter.Talents.SavageStrikes) * 10 * core.CritRatingPerCritChance,
+		CritDamageBonus: hunter.mortalShots(),
+		DamageMultiplier: 1,
```
kayla-glick, line 47: Missing `ThreatMultiplier` here

sim/hunter/wyvern_strike.go
```diff
@@ -42,6 +42,7 @@ func (hunter *Hunter) getWyvernStrikeConfig(rank int) core.SpellConfig {
 			return hunter.DistanceFromTarget <= 5
 		},
 
+		CritDamageBonus: hunter.mortalShots(),
 		DamageMultiplier: 1,
```
kayla-glick, line 46: I missed this before but Wyvern Strike is also missing a threat multiplier. I'm not sure about Wing Clip above? If it deals damage now it probably needs one too

sim/hunter/mongoose_bite.go
```diff
@@ -0,0 +1,110 @@
...
+		MaxStacks: 20,
+		OnReset: func(aura *core.Aura, sim *core.Simulation) {
+			aura.Activate(sim)
+		},
+		OnSpellHitDealt: func(aura *core.Aura, sim *core.Simulation, spell *core.Spell, result *core.SpellResult) {
+			if result.Outcome.Matches(core.OutcomeDodge) {
+				aura.SetStacks(sim, 1);
+				hunter.DefensiveState.Activate(sim)
+			} else if result.Outcome.Matches(core.OutcomeLanded) {
+				if spell.ProcMask == core.ProcMaskMeleeMHAuto || spell.ProcMask == core.ProcMaskMeleeOHAuto {
```
kayla-glick, line 78:
```suggestion
				if spell.Procmask.Matches(ProcMaskMeleeWhiteHit) {
```

sim/hunter/mongoose_bite.go
```diff
@@ -0,0 +1,110 @@
...
+
+		ApplyEffects: func(sim *core.Simulation, target *core.Unit, spell *core.Spell) {
+			hunter.DefensiveState.Deactivate(sim)
+
+			multiplier := 1.0
+			if stacks := hunter.RaptorFuryAura.GetStacks(); stacks > 0 {
+				multiplier *= 1 + raptorFuryDmgMult*float64(stacks)
+			}
+			
+			damage := multiplier * (baseDamage + (hunter.GetStat(stats.AttackPower) * 0.4))
```
kayla-glick, line 57: `+ (hunter.GetStat(stats.AttackPower) * 0.4)`

Is this handling this bit?
> Mongoose Bite also deals additional damage equal to 40% of your Attack Power.

If so we should add it only the hunter has the Cobra Strikes rune equipped (part of separating the rune from the Mongoose Bite spell)

kayla-glick (changes requested):
Left a couple of things to change but nice work!

sim/hunter/mongoose_bite.go
ncberman: Hmm it does say that but that's not how it works on the current round of PTR, It looks like that wording probably comes from this set bonus: https://www.wowhead.com/classic/spell=456394/s03-item-t1-hunter-melee-6p-bonus

sim/hunter/mongoose_bite.go
ncberman: Added!

sim/hunter/wyvern_strike.go
ncberman: Added

sim/hunter/mongoose_bite.go
ncberman: Added

sim/hunter/mongoose_bite.go
ncberman: GetStacks() has a built-in null check but I agree that what I wrote is a bad practice so I implemented a rune-check beforehand

sim/hunter/mongoose_bite.go
ncberman: Hunter tank dream prayge, but I agree it's probably better to implement the spell correctly rather than quickly!

sim/hunter/mongoose_bite.go
kayla-glick: Ooooh ok

kayla-glick (approved):

adocoD:
Friendly reminder that the checklist states what gets affected by each talent. Did all the research beforehand.
But at the same time I also completely overlooked it myself.

PR #849, opened 2024-07-02
Added Kill Shot rune spell for hunters + Explosive Trap / Immolation Trap fix

ncberman:
Kill Shot spell implementation
https://www.wowhead.com/classic/spell=409593/kill-shot#modified-by

Some notes for this spell
Kill Shot is NOT affected by the efficiency talent
Kill Shot DOES count as a shot ability for 'Cobra Strikes' and 'Lock n Load'
Wowhead says weapon damage rather than normalized weapon damage
The %hp reset mechanic is tied to the cast rather than the actual result

#791


sim/hunter/kill_shot.go
```diff
@@ -0,0 +1,76 @@
...
+	"github.com/wowsims/sod/sim/core"
+	"github.com/wowsims/sod/sim/core/proto"
+)
+
+func (hunter *Hunter) registerKillShotSpell() {
+	if !hunter.HasRune(proto.HunterRune_RuneLegsKillShot) {
+		return
+	}
+
+	baseDamage := 113 / 100 * 100 / 100 * (2.976264 + 0.641066 * float64(hunter.Level) + 0.022519 * float64(hunter.Level) * float64(hunter.Level)) // taken from wowhead
```
kayla-glick, line 15: You can replace the base calculation that uses the hunter's level with `hunter.baseRuneAbilityDamage()`

kayla-glick: Also even though the tooltip shows `113 / 100 * 100 / 100 *` we can just do `113 / 100 *`

sim/hunter/kill_shot.go
```diff
@@ -0,0 +1,76 @@
...
+	hasCobraStrikes := hunter.pet != nil && hunter.HasRune(proto.HunterRune_RuneChestCobraStrikes)
+
+	// Efficiency talent doesn't apply to this spell even though it has 'shot' in the name
+	manaCostMultiplier := 1.0
+	if hunter.HasRune(proto.HunterRune_RuneChestMasterMarksman) {
+		manaCostMultiplier -= 0.25
+	}
+
+	hunter.KillShot = hunter.RegisterSpell(core.SpellConfig{
+		ActionID:     core.ActionID{SpellID: 409593},
```
kayla-glick, line 26: For the ID on this spell you can use `int32(proto.HunterRune_RuneLegsKillShot)`

sim/hunter/kill_shot.go
```diff
@@ -0,0 +1,76 @@
...
+
+		ManaCost: core.ManaCostOptions{
+			BaseCost:   0.03,
+			Multiplier: manaCostMultiplier,
+		},
+		Cast: core.CastConfig{
+			DefaultCast: core.Cast{
+				GCD: core.GCDDefault,
+			},
+			IgnoreHaste: true, // Hunter GCD is locked at 1.5s
```
kayla-glick, line 42: This shouldn't actually matter. We removed DoT and GCD haste contribution from Wrath. TBH these options could probably be removed but I figure keeping them around is safer just in case

ncberman: Do you want me to remove them or just leave them in?

sim/hunter/kill_shot.go
ncberman: I saw that baseRuneAbilityDamage function at some point and totally didn't realize it was the same!

kayla-glick (approved):

PR #876, opened 2024-07-07
Updates formula for Blood Fury to include agility AP for rogues

sanguinerarogue:


sanguinerarogue:
Let me know if this works for you @kayla-glick just added rogue AP scaling vs agility.  No druids can be orc so its only rogues that are affected but built it out anyway.

PR #877, opened 2024-07-07
Updated Base ranged range and MS

adocoD:
Hunter melee and ranged only had a difference of 3 yds. It should actually be 7.

Hunter range to boss gets calculated from edge of hitbox. Since melee range in the sim is 5 (4+1), ranged would be 12 (4+8).

Movement speed was also incorrect , so changed it to 7yd per second.


sim/core/unit.go
```diff
@@ -428,7 +428,7 @@ func (unit *Unit) MoveTo(moveRange float64, sim *Simulation) {
 	unit.moveSpell.Cast(sim, unit.CurrentTarget)
 
 	sim.AddPendingAction(NewPeriodicAction(sim, PeriodicActionOptions{
-		Period:          time.Millisecond * 500,
+		Period:          time.Millisecond * 143,
```
kayla-glick, line 431: Why 143ms? That's an oddly specific number 😄

sim/hunter/aimed_shot.go
```diff
@@ -56,7 +56,7 @@ func (hunter *Hunter) getAimedShotConfig(rank int, timer *core.Timer) core.Spell
 			},
 		},
 		ExtraCastCondition: func(sim *core.Simulation, target *core.Unit) bool {
-			return hunter.DistanceFromTarget >= 8
+			return hunter.DistanceFromTarget >= 12
```
kayla-glick, line 59: Maybe we could pull this out into a constant value in `sim/core/constants`? `core.MinRangedAttackDistance`

sim/core/unit.go
adocoD: Time per yd moved. I suppose I could've written 1000 / 7 there to be accurate, since it's 7 yd/s.

adocoD: Fixed this

sim/hunter/aimed_shot.go
adocoD: Added constant for both MaxMeleeAttackDistance and MinRangedAttackDistance

PR #884, opened 2024-07-08
Add all ret runes p4

Adamrch:
RV rune implemented (Used ignite aura as inspiration and modified)
AM rune implemented (adds like 10 dps LOL)
Added aura selection for your primary aura (to work with AM)
Tested RV extensively, but feel free to test more


sim/paladin/paladin.go
```diff
@@ -12,6 +12,7 @@ import (
 var TalentTreeSizes = [3]int{14, 15, 15}
 
 const (
+        SpellFlag_RV = core.SpellFlagAgentReserved1
```
kayla-glick, line 15: Looks like the indentation is a bit off here

sim/paladin/paladin.go
```diff
@@ -77,18 +81,27 @@ func (paladin *Paladin) Initialize() {
 	paladin.registerHammerOfWrath()
 	paladin.registerHolyWrath()
 	paladin.registerAvengingWrath()
+	paladin.registerAuraMastery()
+
+        if paladin.primaryPaladinAura == proto.PaladinAura_SanctityAura {
+            paladin.currentPaladinAura = core.SanctityAuraAura(paladin.GetCharacter())
+        }
```
kayla-glick, line 88: Looks like the indentation is a bit off here too

kayla-glick: Also I think we should go ahead and do this inside the `Reset()` similar to how seals are handled. That way if there's ever a reason to add the ability to switch seals it can be reset between each iteration

sim/paladin/paladin.go
```diff
@@ -77,18 +81,27 @@ func (paladin *Paladin) Initialize() {
...
 	paladin := &Paladin{
 		Character: *character,
 		Talents:   &proto.PaladinTalents{},
 	}
-	core.FillTalentsProto(paladin.Talents.ProtoReflect(), talentsStr, TalentTreeSizes)
+	core.FillTalentsProto(paladin.Talents.ProtoReflect(), options.TalentsString, TalentTreeSizes)
+
+        if pallyAura == proto.PaladinAura_SanctityAura {
+            paladin.primaryPaladinAura = pallyAura
+        }
```
kayla-glick, line 104: Looks like the indentation is a bit off here too

sim/paladin/righteous_vengeance.go
```diff
@@ -0,0 +1,72 @@
...
+				paladin.procRV(sim, result)
+			}
+		},
+	})
+
+	paladin.rv = paladin.RegisterSpell(core.SpellConfig{
+		ActionID:    core.ActionID{SpellID: 440675},
+		SpellSchool: core.SpellSchoolHoly,
+		//DefenseType: core.DefenseTypeMagic,
+		ProcMask:    core.ProcMaskEmpty,  //Not sure
```
kayla-glick, line 35: As long as the RV dot ticks don't proc effects then this is correct

sim/paladin/righteous_vengeance.go
```diff
@@ -0,0 +1,72 @@
...
+		Flags:       core.SpellFlagPureDot | core.SpellFlagIgnoreModifiers | core.SpellFlagIgnoreResists,
```
kayla-glick, line 36: Is the RV DoT guaranteed to be exactly 50% of the original value?

sim/paladin/righteous_vengeance.go
```diff
@@ -0,0 +1,72 @@
...
+	})
+}
+
+func (paladin *Paladin) procRV(sim *core.Simulation, result *core.SpellResult) {
+	dot := paladin.rv.Dot(result.Target)
+
+	newDamage := result.Damage * 0.5
+	outstandingDamage := core.TernaryFloat64(dot.IsActive(), dot.SnapshotBaseDamage*float64(dot.NumberOfTicks-dot.TickCount), 0)
+
+	dot.Snapshot(result.Target, (outstandingDamage+newDamage)/float64(RVTicks), false)
```
kayla-glick, line 69: Do we have confirmation that it works like ignite?

sim/paladin/soc.go
```diff
@@ -90,7 +90,7 @@ func (paladin *Paladin) registerSealOfCommand() {
 			SpellSchool: core.SpellSchoolHoly,
 			DefenseType: core.DefenseTypeMelee,
 			ProcMask:    core.ProcMaskMeleeMHSpecial | core.ProcMaskProc,
-			Flags:       core.SpellFlagMeleeMetrics,
+			Flags:       core.SpellFlagMeleeMetrics | SpellFlag_RV, // RV Worked on PTR
```
kayla-glick, line 93: What exactly does RV do? Looking at the wowhead entry it doesn't seem like seals should proc it?
image

sim/paladin/aura_mastery.go
```diff
@@ -0,0 +1,56 @@
...
+	}
+
+	cd := core.Cooldown{
+		Timer:    paladin.NewTimer(),
+		Duration: time.Minute * 2,
+	}
+
+	aura := paladin.RegisterAura(core.Aura{
+		Label:    "Aura Mastery",
+		ActionID: core.ActionID{SpellID: 415756},
```
kayla-glick, line 24: This spell ID should be 407624, but you can also use `int32(proto.PaladinRune_RuneLegsAuraMastery)` since that has the same value

sim/paladin/aura_mastery.go
```diff
@@ -0,0 +1,56 @@
...
+                    }
+		},
+		OnExpire: func(aura *core.Aura, sim *core.Simulation) {
+                    if paladin.currentPaladinAura != nil {
+                        if paladin.currentPaladinAura.Label == "Sanctity Aura" {
+                            paladin.PseudoStats.SchoolDamageDealtMultiplier[stats.SchoolIndexHoly] /= (1.2 / 1.1)
+                        }
+                    }
+		},
+	})
```
kayla-glick, line 40: Looks like the indentation got a bit wonky here

sim/paladin/paladin.go
Adamrch: I can add that

sim/paladin/paladin.go
Adamrch: will fix

sim/paladin/righteous_vengeance.go
Adamrch: It might be slightly more nuanced than that possibly, but for the most part seem yes it works like that to avoid double scaling with multiplier like Avenging Wrath possibly (caster side) or target side (nightfall / occult poison) 
image

Adamrch: I went off of Ignore damage taken Mod. and ignore caster damage mod

sim/paladin/righteous_vengeance.go
Adamrch: Seems to match ignite behavior to me. https://sod.warcraftlogs.com/reports/8F6JnPajkvZ9YVDG#fight=20&type=damage&source=99&view=events&start=4792427&end=4821193
Look at 33.994 (First Crit) - 101.051 (Last RV tick)

00:33.994 Amereyn Judgement of Martyrdom Testwerk *2604*
00:39.152 Amereyn Crusader Strike Testwerk *1917*
00:48.433	Amereyn Divine Storm Testwerk *1931*
00:53.014	Amereyn Crusader Strike Testwerk *1933*
Total Crit damage 8385

Total Dot Damage 5039 (Including partial resists) , 
image

Only thing is the modifier flags seem questionable because 5039 is > 0.5 * 8350.
So if anything this is a least optimistic implementation possibly?
No AW used during this period.

sim/paladin/soc.go
Adamrch: Seal of Command was added to the abilities that procced RV on the last week of the PTR as an undocumented change. Several Light Club members tested it to verify that is does proc it including myself.  This change was done after RV was functioning like ignite and rolling damage (pooling) so it seem unlikely it was an oversight. If it doesn't make it live we can always remove this flag. I suspect it was a late balancing change to bring Twisting up to par with fast 2H ret.

sim/paladin/aura_mastery.go
Adamrch: will fix

sim/paladin/aura_mastery.go
Adamrch: will fix

sim/paladin/righteous_vengeance.go
Adamrch: I'm not sure if it can proc spell/dot procs, it does have the suppress weapon procs flag though. What proc mask would be best here?

sim/paladin/paladin.go
Adamrch: will fix

sim/paladin/righteous_vengeance.go
Adamrch: I thought so but in the comment below even though the ignite behavior matches, the scaling of the damage seems off. 5039 (actual) / 4191 (expected) = 1.202

Possible Explanations? well Occult poison is Applied and so is Sanctity aura.  (which adds up to 20%)
but so is Vengeance, which adds 15%

I'm not sure what flags would make it match.
Any Suggestions?

Lastly:
https://www.wowhead.com/classic/spell=440675/righteous-vengeance has the no partial resists flag and cannot miss flag but the logs indicate partial resists.

sim/paladin/righteous_vengeance.go
Adamrch: Fixed , though we don't really have many important procs that would proc here I think

sim/paladin/righteous_vengeance.go
Adamrch: resolved by remove resits flag and setting defense type

sim/paladin/paladin.go
```diff
@@ -77,18 +81,27 @@ func (paladin *Paladin) Initialize() {
 	paladin.registerHammerOfWrath()
 	paladin.registerHolyWrath()
 	paladin.registerAvengingWrath()
+	paladin.registerAuraMastery()
+
+	if paladin.primaryPaladinAura == proto.PaladinAura_SanctityAura {
+		paladin.currentPaladinAura = core.SanctityAuraAura(paladin.GetCharacter())
+	}
```
kayla-glick, line 88: I think this shouldn't be needed anymore after adding the `ResetCurrentPaladinAura` bit below

sim/paladin/righteous_vengeance.go
```diff
@@ -0,0 +1,75 @@
...
+			if spell.Flags.Matches(SpellFlag_RV) && result.DidCrit() {
+				paladin.procRV(sim, result)
+			}
+		},
+	})
+
+	paladin.rv = paladin.RegisterSpell(core.SpellConfig{
+		ActionID:    core.ActionID{SpellID: 440675},
+		SpellSchool: core.SpellSchoolHoly,
+		DefenseType: core.DefenseTypeMelee,
```
kayla-glick, line 34: I could be wrong but I think this should maybe be `DefenseTypeMagic`. That will cause it to scale with magic / holy damage taken effects whereas right now I think this might scale with physical damage taken debuffs

sim/paladin/righteous_vengeance.go
```diff
@@ -0,0 +1,75 @@
...
+
+		// SpellFlagIgnoreTargetModifiers was thought to be used based on wowhead flags
+		// WCL parses show that this is not the case
+
+		DamageMultiplier: 1,
+		ThreatMultiplier: 1,
+
+		Cast: core.CastConfig{
+			IgnoreHaste: true,
+		},
```
kayla-glick, line 46: This shouldm't be needed since it doesn't have a cast time

kayla-glick (approved):

sim/paladin/righteous_vengeance.go
Adamrch: Ignite has it, I think it has to do with the Dot Not ticking faster with haste

sim/paladin/righteous_vengeance.go
Adamrch: image

sim/paladin/righteous_vengeance.go
kayla-glick: Oh oops. AFAIK all dots don't tick faster with haste in vanilla. The only thing that field affects is cast time
https://github.com/wowsims/sod/blob/a0e8093b28521bfd54d65ffffabc7fa0ac117162/sim/core/cast.go#L162-L166

So it's probably just a leftover from removing that stuff but no worries!

PR #887, opened 2024-07-08
Fan of Knives and Crimson fixes

sanguinerarogue:
Take a look and let me know.


sim/rogue/runes.go
```diff
@@ -83,10 +83,15 @@ func (rogue *Rogue) applyFocusedAttacks() {
 			aura.Activate(sim)
 		},
 		OnSpellHitDealt: func(aura *core.Aura, sim *core.Simulation, spell *core.Spell, result *core.SpellResult) {
-			if !spell.ProcMask.Matches(core.ProcMaskMeleeOrRanged) || !result.DidCrit() {
+			isFoKOH := false
+
+			if spell.ActionID.SpellID == 409240 && spell.ActionID.Tag == 2 {
+				isFoKOH = true
+			}
```
kayla-glick, line 90: Rather than do this, in the FOK code you could assign the FOK MH/OH spells to the rogue as properties, then just do `spell == rogue.FanOfKnivesOH`

sim/rogue/crimson_tempest.go
```diff
@@ -92,5 +92,8 @@ func (rogue *Rogue) registerCrimsonTempestSpell() {
 }
 
 func (rogue *Rogue) CrimsonTempestDamage(comboPoints int32) float64 {
-	return []float64{0.15, 0.3, 0.45, 0.6, 0.75, 0.9}[comboPoints] * rogue.GetStat(stats.AttackPower)
+    tickDamageValues := []float64{0, 0.3, 0.45, 0.6, 0.75, 0.9}
+    tickDamage := tickDamageValues[comboPoints] * rogue.GetStat(stats.AttackPower)/float64(comboPoints+2)
```
kayla-glick, line 96: > /float64(comboPoints+2)

Should this be comboPoints +1? A 5 combo point Crimson Tempest ticks 6 times right?

kayla-glick: You could probably also pass the number of ticks in from the DoT struct when calling this

PR #900, opened 2024-07-10
model p4 double judge

Adamrch:
This models the behavior from PTR that was also found to be the case on Live servers for p4.
This behavior models double judging with and without the t1 6pc tier set bonus.
It is accurate for those without the bonus.
For the 6pc bonus the only thing that will need to change is the linger time of seals from 0.4 to 6.0 seconds


Adamrch:
Timeline comfirmed JoC janky spellhit calculation (on Judge dummy spell) is working as before. Double judge is working too. 
image

Adamrch:
Judge dummy spell, spellhit roll jankiness intact 
image

Adamrch:
APL used to get Judge to cast during GCD after a twist while both seals are still up. 
image
it seems a strict sequence is required, because for some reason you can't cast judge from the priority list during GCD.

Adamrch:
image
here's without the SoC proccing

Adamrch:
Here's with a full twist  Auto+SoC+SoM+SoM+JoC+JoM 
image

Adamrch:
image

sim/paladin/judgement.go
```diff
@@ -36,13 +36,43 @@ func (paladin *Paladin) registerJudgement() {
...
+				if sealAura.IsActive() {
+				        spell.CalcAndDealOutcome(sim, target, spell.OutcomeMagicHit)
+					//paladin.spellsJoC[i].Cast(sim, target)
+					sealAura.Deactivate(sim)
+				}
+			}
+			
+			for i, sealAura := range paladin.aurasSotC {
+				if sealAura.IsActive() {
+					paladin.spellsJotC[i].Cast(sim, target)
```
kayla-glick, line 59: Do we potentially need to do a `nil` check on these for lower levels where some ranks may not be known?

Adamrch: I just tried running the sim at level 25 and 40 and saw no issues. I can try reviewing the code to see if it is needed.

Adamrch: Though reviewing the code, I don't why it doesn't seem to cause issues. I can add a nil check to each of them.

PR #908, opened 2024-07-17
add ret t1 set bonuses

Adamrch:
Adds t1 ret set bonuses. 
No triple twisting/ judging implemented as it is unknown if that will work.


sim/paladin/paladin.go
```diff
@@ -91,6 +93,13 @@ func (paladin *Paladin) Initialize() {
 	paladin.registerAvengingWrath()
 	paladin.registerAuraMastery()
 
+        paladin.t1SetBonus6Pc = false
+        for _, bonus := range paladin.GetCharacter().GetActiveSetBonusNames() {
+        	if bonus == "Lawbringer Radiance (6pc)" {
+        	    paladin.t1SetBonus6Pc = true
+        	}
+        }
```
kayla-glick, line 101: This can be done with `character.HasSetBonus(ItemSetLawbringerRadiance, 6)`

sim/paladin/item_sets_pve.go
```diff
@@ -101,3 +101,21 @@ var ItemSetSoulforgeArmor = core.NewItemSet(core.ItemSet{
 		},
 	},
 })
+
+var ItemSetLawbringerRadiance = core.NewItemSet(core.ItemSet{
+	Name: "Lawbringer Radiance",
+	Bonuses: map[int32]core.ApplyEffect{
+		2: func(agent core.Agent) {
+			// No need to model
+			//(2) Set : Your Judgement of Light and Judgement of Wisdom also grant the effects of Judgement of the Crusader.
```
kayla-glick, line 110: Sorry I'm confused why is there nothing to do for this?

sim/paladin/paladin.go
```diff
@@ -167,9 +176,14 @@ func (paladin *Paladin) getPrimarySealSpell(primarySeal proto.PaladinSeal) *core
 
 func (paladin *Paladin) applySeal(newSeal *core.Aura, judgement *core.Spell, sim *core.Simulation) {
 	const lingerDuration = time.Millisecond * 400
+	const lingerDuration6pcT1 = time.Second * 6
```
kayla-glick, line 179: I wonder if you could set `lingerDuration` as an attribute on the paladin struct, then in the Lawbringer Radiance 6pc definition instead just override that attribute with `time.Second * 6`? That would cut down on the work done here and having to check for the set bonus in `paladin.go`

sim/paladin/item_sets_pve.go
Adamrch: I suppose it can be modeled. I just don't often see any APL's that explicitly cast these in a rotation and just use the raid debuff settings menu.

ui/retribution_paladin/sim.ts
```diff
@@ -140,18 +143,22 @@ const SPEC_CONFIG = registerSpecConfig(Spec.SpecRetributionPaladin, {
 
 	presets: {
 		rotations: [
+			...Presets.APLPresets[Phase.Phase4],
+			PresetUtils.makePresetAPLRotation('P4 Ret-Exodin', APLP4RetExodinJson),
```
kayla-glick, line 147: Can we do this inside the `presets.ts` file along with the other rotation?

Adamrch: We can. I cant put it in the phase array though

kayla-glick: Oh really? Why not?

Adamrch: it's addressed by the phase variable, so I would have to put it in phase 5 right?

PR #913, opened 2024-07-18
Hunter Cloak Runes + Unit Movement adjustment

ncberman:
**Implemented Volley**
- Channeled spell that does damage over time similar to rain of fire or blizzard
https://www.wowhead.com/classic/spell=14295/volley#modified-by

**Implemented Improved Volley Cloak Rune**
- Removes volley cooldown (when the rune is equipped the spell resets itself on cast)
- Lowers mana cost by 50%
- Volley now does bonus damage equal to 3% of ranged AP.
- No longer suffers spell pushback.
https://www.wowhead.com/classic/spell=440520/improved-volley

**Implemented Hit And Run Cloak Rune**
- After casting raptor strike gain 15% movement speed for 8 sec.
_Note: Had to adjust how the MoveTo function worked to handle floating point rounding errors, also added MoveSpeed as an attribute of a unit. MoveTo spell now uses MoveSpeed to see how fast the unit should move. Reworked the tick period on the Move pending action to be the same as the tick period (shouldn't be used to adjust the move speed)_
https://www.wowhead.com/classic/spell=440533/hit-and-run


sim/core/unit.go
```diff
@@ -417,22 +422,25 @@ func (unit *Unit) MoveTo(moveRange float64, sim *Simulation) {
...
 	sim.AddPendingAction(NewPeriodicAction(sim, PeriodicActionOptions{
-		Period:          time.Millisecond * 1000 / 7,
-		NumTicks:        int(moveTicks),
+		Period:          time.Millisecond * 200,
+		NumTicks:        int(math.Ceil(moveTicks)),
 		TickImmediately: false,
 
 		OnAction: func(sim *Simulation) {
+			remainingDistance := math.Abs(moveRange - unit.DistanceFromTarget)
+			if(remainingDistance < moveInterval) {
```
kayla-glick, line 441: 🔨 In Go you don't have to use parens in a condition like you do in JavaScript so this can just be `if remainingDistance < moveInterval {`

kayla-glick, line 441:
```suggestion
			if remainingDistance < moveInterval {
```

ncberman:
My original rewrite of the moveTo system kept running into issues so I just removed some redundant code and replace the hard coded 7 yd/s to use unit movespeed instead.

sim/hunter/volley.go
```diff
@@ -0,0 +1,102 @@
...
+func (hunter *Hunter) getVolleyConfig(rank int) core.SpellConfig {
+	spellId := [4]int32{0, 1510, 14294, 14295}[rank]
+	baseDamage := [4]float64{0, 50, 65, 80}[rank]
+	manaCost := [4]float64{0, 350, 420, 490}[rank]
+	level := [4]int{0, 40, 50, 58}[rank]
+
+	hasImprovedVolley := hunter.HasRune(proto.HunterRune_RuneCloakImprovedVolley)
+
+	manaCostModifer := 1.0
+	manaCostModifer *= 1 - (0.02 * float64(hunter.Talents.Efficiency))
```
kayla-glick, line 34: You could delete the top line here and just have the bottom with `:=`

sim/hunter/volley.go
```diff
@@ -0,0 +1,102 @@
...
+				GCD: core.GCDDefault,
+			},
+			CD: core.Cooldown{
+				Timer:    hunter.NewTimer(),
+				Duration: time.Second * 60,
+			},
+		},
+		ExtraCastCondition: func(sim *core.Simulation, target *core.Unit) bool {
+			return !hunter.Unit.Moving
+		},
```
kayla-glick, line 65: Is this needed if we have the `core.SpellFlagChanneled` flag on it? If not then I think the core should handle this kind of condition

kayla-glick (approved):

PR #914, opened 2024-07-18
Totem of Rage

tyler-cb:
- Adds Totem of Rage; provides 30 spellpower to each shock while equipped
- Changed the Totem of Rage within the database to the vanilla version, as the version added in SoD appears to be unusued and has no equip effect


sim/shaman/earth_shock.go
```diff
@@ -54,8 +54,10 @@ func (shaman *Shaman) newEarthShockSpellConfig(rank int, shockTimer *core.Timer)
 	spell.ThreatMultiplier = 2
 	spell.BonusCoefficient = spellCoeff
 
+	bonusDamage := core.TernaryFloat64(shaman.Ranged().ID == TotemOfRage, 30*spellCoeff, 0)
```
kayla-glick, line 57: Can you add the bonus damage up above to the `baseDamageLow` and `baseDamageHigh` variables? It should have the same effect but will require 1 less calculation per cast. Same for frost shock

PR #920, opened 2024-07-20
Single Minded Fury reset fix

FelixPflaum:
Resets warrior.lastMeleeAutoTarget back to nil on sim reset. Causes strange(?) stat weight test results. Not sure if this is fine.


kayla-glick (approved):
LGTM. The negative weight is for hit which makes sense if a set is hit capped

PR #921, opened 2024-07-20
Hunter whistle, PvP gloves

adocoD:
So I managed to implement the boots pet damage and the Whistle trinket.

Please look into the whistle code since I couldn't figure out the correct way to call on the current pet.


adocoD:
Removed the boots as they apparently arent ingame.
Whistle moved to hunter items. Easier to get in there and found the other trinket code.
Added the Damage buffs on the pvp gloves, doesnt seem to work yet.

sim/hunter/raptor_strike.go
```diff
@@ -114,6 +114,7 @@ func (hunter *Hunter) newRaptorStrikeHitSpell(rank int, isMH bool) *core.Spell {
 	}
 
 	return hunter.RegisterSpell(core.SpellConfig{
+		SpellCode:   SpellCode_HunterRaptorStrike,
```
kayla-glick, line 117: I think you probably also want to apply this to the queued spell below

kayla-glick: Actually you may want to do what I did in other sims and separate out spell codes for the base spell, the MH damage, the OH damage, and the queued version
```
SpellCode_HunterRaptorStrike
SpellCode_HunterRaptorStrikeMH
SpellCode_HunterRaptorStrikeOH
SpellCode_HunterRaptorStrikeQueued
```

sim/hunter/items.go
```diff
@@ -148,6 +153,124 @@ func init() {
...
+			Flags:    core.SpellFlagNoOnCastComplete | core.SpellFlagOffensiveEquipment,
+
+			Cast: core.CastConfig{
+				CD: core.Cooldown{
+					Timer:    hunter.NewTimer(),
+					Duration: time.Minute * 1,
+				},
+			},
+			ExtraCastCondition: func(sim *core.Simulation, target *core.Unit) bool {
+				return hunter.pet.IsEnabled()
```
kayla-glick, line 219: Will this potentially error if no pet is selected?

kayla-glick: Or I wonder do you need this cast condition with the `ShouldActivate` on the major cooldown?

sim/hunter/raptor_strike.go
adocoD: I kinda knew this might be commented, but is it necessary when that one line currently increases damage for all RS versions?

Added it to the queued for added security. But is the MH OH split needed?

sim/hunter/items.go
adocoD: Simple fix was just changing the first lines to if return on ==nil

sim/hunter/raptor_strike.go
adocoD: I did extensively test it btw

kayla-glick: Oh it really does it for all of them? Ok I didn't realize that

kayla-glick (approved):

sim/hunter/raptor_strike.go
adocoD: Yeh no clue what magic causes this, but the math maths in tests

PR #923, opened 2024-07-22
Adjust Pet Rotation to ignore Special Abilities when using Flanking Strike Rune

ncberman:
The Flanking Strike rune only resets when on pet abilities, since you want Flanking Strike available as much as possible you always want to turn off higher focus cost abilities and only use focus dump abilities to reset Flanking Strike more often.


sim/hunter/pet.go
```diff
@@ -247,7 +247,7 @@ func (hp *HunterPet) ExecuteCustomRotation(sim *core.Simulation) {
 			_ = tryCast(hp.focusDump) || tryCast(hp.specialAbility)
 		}
 	} else {
-		if hp.specialAbility.IsReady(sim) {
+		if hp.specialAbility.IsReady(sim) && !hp.hunterOwner.HasRune(proto.HunterRune_RuneLegsFlankingStrike) {
```
kayla-glick, line 250: 🔨 Can we cache this as an attribute on the pet? Checking this every time will be fairly expensive

ncberman: Great catch, looking back at the code I didn't realize that the pet actually has its own version of the spell when the flanking strike rune is equipped. I changed it to only use special abilities if the flanking strike spell is null on the pet.

kayla-glick (approved):

PR #924, opened 2024-07-22
Blazefury flametongue

tyler-cb:
Blazefury medallion's fire strike is causing an additional flametongue weapon proc when the FT imbued weapon strikes.

image

image

### DIFF
--- a/sim/core/flags.go
+++ b/sim/core/flags.go
@@ -55,7 +55,8 @@ const (
 	// Mask for Seal of Righteousness, it does not proc Wild Strikes
 	ProcMaskSupressExtraAttack
 	// Mask for Fiery Weapon and Blazefury Medalion that trigger melee procs like Art of War Rune or Vengeance Talent
-	ProcMaskTriggerInstant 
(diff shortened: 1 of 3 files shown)


sim/core/flags.go
```diff
@@ -55,7 +55,9 @@ const (
 	// Mask for Seal of Righteousness, it does not proc Wild Strikes
 	ProcMaskSupressExtraAttack
 	// Mask for Fiery Weapon and Blazefury Medalion that trigger melee procs like Art of War Rune or Vengeance Talent
-	ProcMaskTriggerInstant 
+	ProcMaskTriggerInstant
+	// Mask for Blazefury Medallion triggering on offhand
+	ProcMaskTriggerInstantOH
```
kayla-glick, line 60: I don't think we should just be creatingg proc masks like this. I already want to go back and refactor `ProcMaskTriggerInstant` and `ProcMaskSupressExtraAttack` to not be proc masks because I feel like these go against what a proc mask is used for.

PR #991, opened 2024-08-21
Add target swap to dot portion of flame shock logic

Foknetics:


sim/shaman/flame_shock.go
```diff
@@ -119,15 +119,15 @@ func (shaman *Shaman) newFlameShockSpell(rank int, shockTimer *core.Timer) core.
...
-
 				if shaman.HasRune(proto.ShamanRune_RuneLegsAncestralGuidance) {
 					shaman.lastFlameShockTarget = target
 				}
 
 				if hasPowerSurgeRune && sim.Proc(ShamanPowerSurgeProcChance, "Power Surge Proc") {
 					shaman.PowerSurgeAura.Activate(sim)
 				}
+
+				target = sim.Environment.NextTargetUnit(target)
```
kayla-glick, line 130: I think the way I would do this is actually apply the DoT on line 121 to `result.Target` instead of `target`, because that should be the correct target to apply the DoT to

PR #998, opened 2024-09-01
Hunter Phase 5 Tier

ncberman:
Ranged T2 Tier - https://www.wowhead.com/classic/item-set=1805/dragonstalkers-pursuit
Melee T2 Tier - https://www.wowhead.com/classic/item-set=1806/dragonstalkers-prowess
Predator's Armor - https://www.wowhead.com/classic/item-set=1825/predators-armor

Also added an 8s ICD to lock and load and refactored lock and load rune to utilize SpellFlagTrap
Had to create some new fields to track hunter spell types such as Strikes and Melee abilities to make tier set bonuses easier, also introduced new Pseudostat called MHDpsMultiplier to accommodate melee t2 hunter 4pc bonus


sim/hunter/serpent_sting.go
```diff
@@ -93,7 +100,11 @@ func (hunter *Hunter) chimeraShotSerpentStingSpell(rank int) *core.Spell {
 		BonusCoefficient:         0.4,
 
 		ApplyEffects: func(sim *core.Simulation, target *core.Unit, spell *core.Spell) {
-			spell.CalcAndDealDamage(sim, target, baseDamage, spell.OutcomeRangedCritOnly)
+			damage := baseDamage
+			if has6PDragonstalkerPursuit {
+				damage = damage + (0.25 * hunter.GetStat(stats.RangedAttackPower)) / 5
```
kayla-glick, line 105: The way this type of scaling should be applied is with `spell.RangedAttackPower(target)` instead of just getting the stat directly

sim/hunter/item_sets_pve.go
```diff
@@ -191,3 +191,170 @@ var ItemSetGiantstalkerPursuit = core.NewItemSet(core.ItemSet{
...
+					}
+				},
+			})
+
+			core.MakePermanent(hunter.RegisterAura(core.Aura{
+				Label: "S03 - Item - T2 - Hunter - Melee 2P Bonus Trigger",
+				OnCastComplete: func(aura *core.Aura, sim *core.Simulation, spell *core.Spell) {
+					if spell == hunter.RaptorStrike {
+						procAura.Activate(sim)
+					}
```
kayla-glick, line 237: Instead of checking `spell == X` we should generally compare the `spell.SpellCode` because it's quicker and more reliable than comparing against the whole spell struct. This also wouldn't work for ranked spells where multiple are available to cast

sim/core/stats/stats.go
```diff
@@ -418,6 +418,7 @@ type PseudoStats struct {
 	//  This includes almost all "(Normalized) Weapon Damage", but also some "School Damage (Physical)" abilities.
 	BonusDamage float64 // Comes from '+X Weapon Damage' effects
 
+	MHDpsMultiplier float64 // Used by Hunter T2 Melee 4pc
```
kayla-glick, line 421: We should set the default multiplier to 1 inside the `NewPseudostats` function, then update the `attack.go` code to just use the multiplier and not add 1

sim/core/spell_result.go
```diff
@@ -494,7 +494,6 @@ func (spell *Spell) DealPeriodicHealing(sim *Simulation, result *SpellResult) {
 
 func (spell *Spell) CalcAndDealHealing(sim *Simulation, target *Unit, baseHealing float64, outcomeApplier OutcomeApplier) *SpellResult {
 	result := spell.CalcHealing(sim, target, baseHealing, outcomeApplier)
-	fmt.Println(sim.CurrentTime, spell.ActionID, result.Damage, result.Outcome)
```
kayla-glick, line 497: Oh woops... that was me

kayla-glick: That's a very expensive print 😬

kayla-glick: Ended up just hotfixing this out on master

sim/hunter/talents.go
```diff
@@ -116,11 +116,11 @@ func (hunter *Hunter) registerBestialWrathCD() {
 	}
 
 	actionID := core.ActionID{SpellID: 19574}
-
+	
 	bestialWrathPetAura := hunter.pet.RegisterAura(core.Aura{
 		Label:    "Bestial Wrath Pet",
 		ActionID: actionID,
-		Duration: time.Second * 18,
+		Duration: time.Second * core.TernaryDuration(core.HasItemEffect(MaelstromsWrath), 21, 18),
```
kayla-glick, line 123: This should be able to be done in the item effect definition by using a dummy aura with an `OnInit` callback. You would need to make this aura available through the hunter struct to do that

sim/hunter/items.go
```diff
@@ -273,6 +277,46 @@ func init() {
...
 	})
+
+	core.NewItemEffect(MaelstromsWrath, func(a core.Agent) {
+		hunter := a.(HunterAgent).GetHunter()
+
+		if hunter.pet == nil {
+			return
+		}
+
+		hunter.pet.PseudoStats.DamageDealtMultiplier *= 1.02
```
kayla-glick, line 288: Following up on the comment above:

```suggestion
		if hunter.pet == nil {
			return
		}

		hunter.pet.PseudoStats.DamageDealtMultiplier *= 1.02
		
		if !hunter.talents.BestialWrath {
		  return
		}
		
    hunter.RegisterAura(core.Aura{
		  Label: "Improved Bestial Wrath",
		  OnInit: func(aura *core.Aura, sim *core.Simulation) {
		    hunter.bestialWrathPetAura.Duration += 3
		  }
		}))
```

sim/hunter/pet.go
```diff
@@ -188,7 +188,8 @@ func (hp *HunterPet) Initialize() {
 	hp.focusDump = hp.NewPetAbility(hp.config.FocusDump, false)
 
 	focusRegenMultiplier := (1.0 + 0.1*float64(hp.hunterOwner.Talents.BestialDiscipline)) *
-		core.TernaryFloat64(hp.hunterOwner.HasRune(proto.HunterRune_RuneHandsBeastmastery), 1.5, 1.0)
+		core.TernaryFloat64(hp.hunterOwner.HasRune(proto.HunterRune_RuneHandsBeastmastery), 1.5, 1.0) *
+		core.TernaryFloat64(hp.hunterOwner.HasPredatorArmor[5], 1.2, 1.0)
```
kayla-glick, line 192: To allow this to be done inside of the item set code, we could expose a method for modifying the focus regen multiplier in `focus.go`. I took a similar approach in the rage code in order to apply some of warrior's dynamic rage multipliers

kayla-glick: That would also allow us to apply the Beast Mastery multiplier inside that rune's code

sim/hunter/pet.go
```diff
@@ -277,7 +278,7 @@ func (hunter *Hunter) makeStatInheritance() core.PetStatInheritance {
 		return stats.Stats{
 			stats.Stamina:     ownerStats[stats.Stamina] * 0.3,
 			stats.Armor:       ownerStats[stats.Armor] * 0.35,
-			stats.AttackPower: ownerStats[stats.RangedAttackPower] * 0.22,
+			stats.AttackPower: ownerStats[stats.RangedAttackPower] * 0.22 * core.TernaryFloat64(hunter.HasPredatorArmor[3], 1.2, 1.0),
```
kayla-glick, line 281: Along similar lines, we could potentially create a function for updating stat inheritance, then in the set bonus code grab the old inheritance and return new inheritance multiplying the old value by the 1.2 multiplier

sim/hunter/item_sets_pve.go
```diff
@@ -191,3 +191,170 @@ var ItemSetGiantstalkerPursuit = core.NewItemSet(core.ItemSet{
...
+		2: func(agent core.Agent) {
+			hunter := agent.(HunterAgent).GetHunter()
+
+			procAura := hunter.RegisterAura(core.Aura{
+				ActionID: core.ActionID{SpellID: 467329},
+				Label:    "S03 - Item - T2 - Hunter - Melee 2P Bonus",
+				Duration: time.Second * 5,
+				OnGain: func(aura *core.Aura, sim *core.Simulation) {
+					for _, spell := range hunter.MeleeSpells {
+						if spell != nil {
```
kayla-glick, line 208: You should be safe to not check for nil here because you use `OnSpellRegistered` to fill this

sim/core/attack.go
```diff
@@ -127,10 +127,10 @@ func (weapon *Weapon) CalculateNormalizedWeaponDamage(sim *Simulation, attackPow
 }
 
 func (unit *Unit) MHWeaponDamage(sim *Simulation, attackPower float64) float64 {
-	return unit.AutoAttacks.mh.CalculateWeaponDamage(sim, attackPower)
+	return unit.AutoAttacks.mh.CalculateWeaponDamage(sim, attackPower) * (1 + unit.PseudoStats.MHDpsMultiplier)
```
kayla-glick, line 130: Now that I think about it, rather than creating a multiplier pseudostat for this, couldn't you just multiply `unit.AutoAttacks.mh.BaseDamageMin` and `unit.AutoAttacks.mh.BaseDamageMax` dynamically?

kayla-glick: Potentially creating a method `(weapon *Weapon) ModWeaponDamage(multiplier float64)` to do this?

sim/hunter/item_sets_pve.go
```diff
@@ -191,3 +191,170 @@ var ItemSetGiantstalkerPursuit = core.NewItemSet(core.ItemSet{
...
+				Label: "S03 - Item - T2 - Hunter - Ranged 2P Bonus",
+				ActionID: core.ActionID{SpellID: 467235},
+				Duration: time.Second * 1, // Duration doesn't really matter as the duration is set to the duration of the trap debuff that applies this aura
+				OnGain: func(aura *core.Aura, sim *core.Simulation) {
+					hunter.AimedShot.DamageMultiplier *= 1.20
+				},
+				OnExpire: func(aura *core.Aura, sim *core.Simulation) {
+					hunter.AimedShot.DamageMultiplier /= 1.20
+				},
+			})
```
kayla-glick, line 288: This won't quite work because it's a general 20% damage without actually checking if the target is afflicted by one of the traps. This is mostly important for immolation trap since it's a single-target trap. I think this probably has to be done inside of the aimed shot code itself unfortunately

sim/hunter/hunter.go
```diff
@@ -89,6 +89,9 @@ type Hunter struct {
 	WingClip       *core.Spell
 
 	Shots []*core.Spell
+	Strikes []*core.Spell
+	MeleeSpells []*core.Spell
+	LastShot *core.Spell
```
kayla-glick, line 94: Since this is set dynamically, you would want to make sure to set it to `nil` in the `Reset` method so that it doesn't leak across iterations

sim/hunter/item_sets_pve.go
```diff
@@ -191,3 +191,170 @@ var ItemSetGiantstalkerPursuit = core.NewItemSet(core.ItemSet{
...
+			hunter := agent.(HunterAgent).GetHunter()
+
+			procAura := hunter.RegisterAura(core.Aura{
+				ActionID: core.ActionID{SpellID: 467312},
+				Label:    "S03 - Item - T2 - Hunter - Ranged 4P Bonus",
+				Duration: time.Second * 12,
+
+				OnGain: func(aura *core.Aura, sim *core.Simulation) {
+					for _, spell := range hunter.Shots {
+						if spell != nil && spell != hunter.LastShot {
```
kayla-glick, line 311: You might be able to not check for not nil here, and then you may also want to check spell codes instead of just the spell struct. You would need to also check for a `nil` LastShot before checking SpellCode though

sim/hunter/item_sets_pve.go
```diff
@@ -191,3 +191,170 @@ var ItemSetGiantstalkerPursuit = core.NewItemSet(core.ItemSet{
...
+						procAura.Deactivate(sim)
+						hunter.LastShot = spell
+						procAura.Activate(sim)
+					}
+				},
+			}))
+		},
+		//  Your Serpent Sting damage is increased by 25% of your Attack Power over its normal duration.
+		6: func(agent core.Agent) {
+			// Implemented in serpent_sting.go
```
kayla-glick, line 338: How you could set this up here is by creating a `serpentStingPctAP` variable that keeps track of the AP scaling of Serpent sting. Set it to 0 by default but always multiply the AP by it inside of the serpent sting code, then override the value to `0.25` here

sim/hunter/item_sets_pve.go
```diff
@@ -191,3 +191,170 @@ var ItemSetGiantstalkerPursuit = core.NewItemSet(core.ItemSet{
...
+						oldStatInheritance := hunter.pet.GetStatInheritance()
+						hunter.pet.UpdateStatInheritance(
+							func(ownerStats stats.Stats) stats.Stats {
+								s := oldStatInheritance(ownerStats)
+								s[stats.AttackPower] *= 1.20
+								return s
+							})
+					}
+				},
+			}))
```
kayla-glick, line 345: Just a little knit but could you avoid building this aura if the pet is nil?

sim/hunter/item_sets_pve.go
```diff
@@ -191,3 +191,170 @@ var ItemSetGiantstalkerPursuit = core.NewItemSet(core.ItemSet{
...
+		5: func(agent core.Agent) {
+			hunter := agent.(HunterAgent).GetHunter()
+			core.MakePermanent(hunter.RegisterAura(core.Aura{
+				Label: "Predator's Armor 5P",
+				OnInit: func(aura *core.Aura, sim *core.Simulation) {
+					if hunter.pet != nil {
+						hunter.pet.AddFocusRegenMultiplier(1.20)
+					}
+				},
+			}))
```
kayla-glick, line 357: Same here

sim/hunter/item_sets_pve.go
```diff
@@ -191,3 +191,170 @@ var ItemSetGiantstalkerPursuit = core.NewItemSet(core.ItemSet{
...
+				OnCastComplete: func(aura *core.Aura, sim *core.Simulation, spell *core.Spell) {
+					if !spell.ProcMask.Matches(core.ProcMaskMeleeMHSpecial) {
+						return
+					}
+
+					aura.Deactivate(sim)
+					if spell.SpellCode == SpellCode_HunterRaptorStrike {
+						aura.Activate(sim)
+					}
+				},
```
kayla-glick, line 225: With the below aura won't this double proc it? It's minor but you could probably just return if the spell is raptor strike and never disable it since it'll be refreshed anyways

sim/hunter/item_sets_pve.go
```diff
@@ -191,3 +191,170 @@ var ItemSetGiantstalkerPursuit = core.NewItemSet(core.ItemSet{
...
+				Label:    "S03 - Item - T2 - Hunter - Ranged 4P Bonus",
+				Duration: time.Second * 12,
+
+				OnGain: func(aura *core.Aura, sim *core.Simulation) {
+					for _, spell := range hunter.Shots {
+						if hunter.LastShot != nil && spell.SpellCode != hunter.LastShot.SpellCode {
+							spell.DamageMultiplier *= 1.10
+						}
+					}
+				},
```
kayla-glick, line 293: In hindsight `LastShot` should always be set by the trigger aura so that nil check shouldn't be needed 😅 sorry about that!

sim/hunter/item_sets_pve.go
```diff
@@ -191,3 +191,170 @@ var ItemSetGiantstalkerPursuit = core.NewItemSet(core.ItemSet{
...
+						procAura.Deactivate(sim)
+						hunter.LastShot = spell
+						procAura.Activate(sim)
+					}
+				},
+			}))
+		},
+		//  Your Serpent Sting damage is increased by 25% of your Attack Power over its normal duration.
+		6: func(agent core.Agent) {
+			agent.(HunterAgent).GetHunter().SerpentStingAPCoeff += 0.25
```
kayla-glick, line 316: Mentioned in DM but I think this needs to be done in an `OnInit` because item code is run before spell code

sim/hunter/item_sets_pve.go
ncberman: Yeah the problem I had was that if raptor strike consumed the buff it wouldn't reactivate it so I manually added it here, I could change it to simply ignore raptor strike instead of doing this reactivation

PR #1000, opened 2024-09-02
P5 Shockadin related changes

mlizard32:
```
-Added Regex String to handle new Hybrid spelldmg/heal stats commonly found on paladin items (judgment set) when building items database 
-Added back Horn of Lordaeron to Buffs menu
-Added Item sets
        -Freethinker's Armor
        -Merciful Judgement
-Added Item effects
	-HerosBrand
	-zandalarFreethinkersBreastplate
	-zandalarFreethinkersBelt
	-grileksCharmOfValor
	-Truthbearer

-Added Shockadin p5 gear set and APL
-Added Shockadin Test
-Removed Gearset/APL older than phase 4 from UI
```


sim/common/sod/item_effects/phase_5.go
```diff
@@ -9,6 +9,8 @@ import (
 const (
 	Stormwrath     = 231387
 	LightningsCell = 231784
+	Truthbearer1H  = 229806
+	Truthbearer2H  = 229749
```
kayla-glick, line 13: 🔨 Tiny knit but can you order there by the ID # in ascending order?

proto/common.proto
```diff
@@ -642,6 +642,7 @@ message RaidBuffs {
 
 	// SoD Runes
 	int32 demonic_pact = 32;
+	bool horn_of_lordaeron = 33;
```
kayla-glick, line 645: I was wrong to ever get rid of it ☹️

sim/common/sod/item_effects/phase_5.go
```diff
@@ -61,6 +63,52 @@ func init() {
...
+		procMask := character.GetProcMaskForItem(Truthbearer2H)
+		crusadersZealAura := crusadersZealAura465414(character)
+
+		core.MakeProcTriggerAura(&character.Unit, core.ProcTrigger{
+			Name:              "Truthbearer (Crusaders Zeal)",
+			Callback:          core.CallbackOnSpellHitDealt,
+			Outcome:           core.OutcomeLanded,
+			ProcMask:          procMask,
+			SpellFlagsExclude: core.SpellFlagSuppressWeaponProcs,
+			PPM:               1, // Estimated based on data from WoW Armaments Discord
```
kayla-glick, line 81: I assume this might be a copy paste but could you remove the comments on these two about the armaments discord since they need to be confirmed?

sim/core/buffs.go
```diff
@@ -765,7 +784,10 @@ func applyBuffEffects(agent Agent, playerFaction proto.Faction, raidBuffs *proto
 	if raidBuffs.BattleShout != proto.TristateEffect_TristateEffectMissing {
 		MakePermanent(BattleShoutAura(&character.Unit, GetTristateValueInt32(raidBuffs.BattleShout, 0, 5), 0))
 	}
-	if individualBuffs.BlessingOfMight != proto.TristateEffect_TristateEffectMissing && isAlliance {
+	
+	if raidBuffs.HornOfLordaeron && isAlliance {
+		character.AddStats(BuffSpellByLevel[HornOfLordaeron][level])
```
kayla-glick, line 789: Could you potentially create an aura for this? It just helps users feel confident that buffs are actually active. A lot of the old buff code still needs to be converted

sim/paladin/item_sets_pve.go
```diff
@@ -120,3 +120,63 @@ var ItemSetLawbringerRadiance = core.NewItemSet(core.ItemSet{
...
+			paladin := agent.GetCharacter()
+			paladin.OnSpellRegistered(func(spell *core.Spell) {
+				if spell.SpellCode == SpellCode_PaladinHolyShock {
+					spell.DamageMultiplier *= 1.5
+				}
+			})
+		},
+		5: func(agent core.Agent) {
+			// Reduce cooldown of Exorcism by 3 seconds
+
```
kayla-glick, line 148: Looks like this implementation got missed?

sim/paladin/items.go
```diff
@@ -16,6 +16,10 @@ const (
 	LibramDiscardedTenetsOfTheSilverHand = 209574
 	LibramOfBenediction                  = 215435
 	LibramOfDraconicDestruction          = 221457
+	HerosBrand							 = 231328
+	zandalarFreethinkersBreastplate      = 231329
+	zandalarFreethinkersBelt             = 231330
+	grileksCharmOfValor                  = 231285
```
kayla-glick, line 22: 🔨 Small knit. Would you mind ordering these by ID descending and then capitalizing the first letter of the variable names to be consistent with others?

sim/paladin/items.go
```diff
@@ -34,4 +38,76 @@ func init() {
...
+	})
+
+
+	// https://www.wowhead.com/classic/item=231285/grileks-charm-of-valor
+	// Use: Increases the critical hit chance of Holy spells by 10% for 15 sec. If Shock and Awe is engraved, gain an additional 5%. (1 Min, 30 Sec Cooldown)
+	core.NewItemEffect(grileksCharmOfValor, func(agent core.Agent) {
+		character := agent.GetCharacter()
+		paladin := agent.(PaladinAgent).GetPaladin()
+
+		character.PseudoStats.BonusDamage += 4
```
kayla-glick, line 71: This might be a copy-paste error? I don't see a bonus damage effect on this trinket

sim/paladin/items.go
```diff
@@ -34,4 +38,76 @@ func init() {
...
+		character := agent.GetCharacter()		
+		character.PseudoStats.SchoolBonusHitChance[stats.SchoolIndexHoly] += 2
+	})
+
+
+	// https://www.wowhead.com/classic/item=231285/grileks-charm-of-valor
+	// Use: Increases the critical hit chance of Holy spells by 10% for 15 sec. If Shock and Awe is engraved, gain an additional 5%. (1 Min, 30 Sec Cooldown)
+	core.NewItemEffect(grileksCharmOfValor, func(agent core.Agent) {
+		character := agent.GetCharacter()
+		paladin := agent.(PaladinAgent).GetPaladin()
```
kayla-glick, line 69: 🔨 Tiny knit. You could just get the paladin here and register everything using it instead of the character

sim/paladin/items.go
```diff
@@ -34,4 +38,76 @@ func init() {
...
+
+		aura := character.RegisterAura(core.Aura{
+			ActionID: core.ActionID{ItemID: grileksCharmOfValor},
+			Label:    "Brilliant Light",
+			Duration: duration,
+			OnGain: func(aura *core.Aura, sim *core.Simulation) {
+				character.AddStatDynamic(sim, stats.SpellCrit, core.TernaryFloat64(paladin.hasRune(proto.PaladinRune_RuneCloakShockAndAwe), 15.0 * core.CritRatingPerCritChance, 10.0 * core.CritRatingPerCritChance))
+			},
+			OnExpire: func(aura *core.Aura, sim *core.Simulation) {
+				character.AddStatDynamic(sim, stats.SpellCrit, core.TernaryFloat64(paladin.hasRune(proto.PaladinRune_RuneCloakShockAndAwe), -15.0 * core.CritRatingPerCritChance, -10.0 * core.CritRatingPerCritChance))
```
kayla-glick, line 83: It looks like the trinket only applies to holy spells. We could either create a school crit multiplier pseudostat or alternatively you could just apply the bonus crit to any of the paladin's spells matching the holy school. It's odd though the wowhead entry looks like it just gives the paladin 10% crit with no specific school

kayla-glick: image

sim/paladin/retribution/retribution_test.go
```diff
@@ -216,6 +233,21 @@ var Phase3Consumes = core.ConsumesCombo{
 		ZanzaBuff:         proto.ZanzaBuff_AtalaiMojoOfWar,
 	},
 }
+var Phase4Consumes = core.ConsumesCombo{
+	Label: "Phase 4 Consumes",
```
kayla-glick, line 237: Small knit, should this be phase 5?

ui/retribution_paladin/sim.ts
```diff
@@ -128,24 +128,27 @@ const SPEC_CONFIG = registerSpecConfig(Spec.SpecRetributionPaladin, {
...
 	presets: {
 		rotations: [
+			...Presets.APLPresets[Phase.Phase5],
 			...Presets.APLPresets[Phase.Phase4],
-			...Presets.APLPresets[Phase.Phase3],
-			...Presets.APLPresets[Phase.Phase2],
-			...Presets.APLPresets[Phase.Phase1],
+			//...Presets.APLPresets[Phase.Phase3],
+			//...Presets.APLPresets[Phase.Phase2],
+			//...Presets.APLPresets[Phase.Phase1],
```
kayla-glick, line 135: We should still leave all of the old presets available for potentially lower level simming

kayla-glick: Since there's SO many paladin presets especially in phase 4, what we can do is apply level restrictions to them. Take a look at https://github.com/wowsims/sod/blob/master/ui/mage/presets.ts if you'd like an example of that. It'll show only relevant presets based on the player's selected level

kayla-glick (commented):
Great job with this! A couple of knits and things to fix but really excited to see what shockadin can do!

ui/retribution_paladin/sim.ts
mlizard32: I did not know that was available, it is much better this way

sim/paladin/retribution/retribution_test.go
mlizard32: I prefer it as phase 4, since I'm not aware of any consumables that were added in phase5

sim/paladin/item_sets_pve.go
kayla-glick: Huh that's weird, I'll pull down the branch and play around with it

Adamrch: you need to refresh / update the aura

kayla-glick: I put up a PR against your branch https://github.com/mlizard32/wowsims_sod/pull/1! I updated the purifying power code a bit then changed the ZG 5-set to an aura with an `OnInit` to make sure it gets applied after the purifying power rune. This means that Exorcism with both will go from 15 / 2 = 7.5 - 3 = 4.5 seconds. I don't know for sure if the game will do that too or if it will do 15 - 3 = 12 / 2 = 6 but I feel like it'll be 4.5 seconds.

The issue was that you have to modify the exorcism spell's cooldown directly because `paladin.exorcismCD` is passed as a value. Really we should be storing the exorcism timer, as that's still a pointer

mlizard32: Sounds good, merged that

kayla-glick (approved):

PR #1001, opened 2024-09-02
Hunter runes, item, spell update

adocoD:
Updated runes, added r12 pvp gloves effect, added Peregrine effect, added falcon.

Currently issues with Falcon and Peregrine, but im not sure if thats because my host refuses to update to new version (feel like this is the case).

Also, the beast mastery and cobra slayer runes arent swapped in UI yet, need to figure that out.

Added as draft, please review.


sim/hunter/aspects.go
```diff
@@ -8,86 +8,188 @@ import (
...
+                    return
+                }
+
+                if impHawkAura != nil && sim.RandomFloat("Imp Aspect of the Hawk") < improvedHawkProcChance {
+                    impHawkAura.Activate(sim)
+                }
+            }
+        })
+
+    aspectOfTheHawkAura.NewExclusiveEffect("Aspect", true, core.ExclusiveEffect{})
```
ncberman, line 90: Don't forget to add this exclusive effect to the aspect of the falcon code as well

sim/hunter/aspects.go
```diff
@@ -8,86 +8,188 @@ import (
...
+        actionID,
+        stats.Stats{stats.RangedAttackPower: rap},
+        core.NeverExpires,
+        func(aura *core.Aura) {
+            aura.OnSpellHitDealt = func(aura *core.Aura, sim *core.Simulation, spell *core.Spell, result *core.SpellResult) {
+                if spell != hunter.AutoAttacks.RangedAuto() {
+                    return
+                }
+
+                if impHawkAura != nil && sim.RandomFloat("Imp Aspect of the Hawk") < improvedHawkProcChance {
```
ncberman, line 84: Generally we should use sim.Proc() rather than sim.RandomFloat() to check for proc chance
(as well as in the falcon code)

ncberman (commented):
A few comments on the aspect and rune code changes, still taking a look at Peregrine and Rank gear stuff

PR #1014, opened 2024-09-04
Add R12/R13 Set for Hunter

Kawney:
This is my second attempt to try and provide to this project; I am still rather inexperienced and unsure of what I'm doing, BUT here I am attempting to add both factions PvP set bonuses, linked here: https://www.wowhead.com/classic/item-set=1725/warlords-pursuit https://www.wowhead.com/classic/item-set=1726/warlords-prowess https://www.wowhead.com/classic/item-set=1739/field-marshals-pursuit https://www.wowhead.com/classic/item-set=1738/field-marshals-prowess

If there's any issues / errors with these changes, assistance may be necessary / extremely appreciated!

### DIFF
--- a/sim/hunter/item_sets_pvp.go
+++ b/sim/hunter/item_sets_pvp.go
@@ -162,19 +162,23 @@ var ItemSetWarlordsPursuit = core.NewItemSet(core.ItemSet{
 var ItemSetWarlordsProwess = core.NewItemSet(core.ItemSet{
 	Name: "Warlord's Prowess",
 	Bonuses: map[int32]core.ApplyEffect{
-		// +20 Agility.
+		// +20 Agility, 20 stamina
 		2: func(agent core.Agent) {
 			c := agent.GetCharacter()
-			c.AddStat(stats.Agility, 20)
+			c.AddStats(stats.Stats{
+				stats.Agility:       20,
+				stats.Stamina:       20,
 		},
 		// Increases the duration of your Wing Clip by 2 sec.
 		4: func(agent core.Agent) {
 			// Nothing to do
 		},
-		// +40 Attack Power.
+		// +40 Agnostic Attack Power
 		6: func(agent core.Agent) {
 			c := agent.GetCharacter()
-			c.AddStat(stats.AttackPower, 40)
+			c.AddStats(stats.Stats{
+				stats.AttackPower:       40,
+				stats.RangedAttackPower: 40,
 		},
 	},
 })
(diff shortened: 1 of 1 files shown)


sim/hunter/item_sets_pvp.go
```diff
@@ -128,3 +128,99 @@ var ItemSetLieutenantCommandersProwess = core.NewItemSet(core.ItemSet{
...
+})
+
+var ItemSetWarlordsProwess = core.NewItemSet(core.ItemSet{
+	Name: "Warlord's Prowess",
+	Bonuses: map[int32]core.ApplyEffect{
+		// +20 Agility.
+		2: func(agent core.Agent) {
+			c := agent.GetCharacter()
+			c.AddStat(stats.Agility, 20)
+		},
```
kayla-glick, line 169: Looks like Warlord's Prowess also increases stam by 20
image

sim/hunter/item_sets_pvp.go
```diff
@@ -128,3 +128,99 @@ var ItemSetLieutenantCommandersProwess = core.NewItemSet(core.ItemSet{
 		},
 	},
 })
+
+///////////////////////////////////////////////////////////////////////////
+//                            SoD Phase 5 Item Sets
+///////////////////////////////////////////////////////////////////////////
+
+var ItemSetWarlordsPursuit = core.NewItemSet(core.ItemSet{
```
kayla-glick, line 136: It looks like the gloves of the Pursuit sets have +4% damage to multi-show as well. That can be done in `sim/hunter/items.go` for both pairs
image

sim/hunter/item_sets_pvp.go
```diff
@@ -128,3 +128,99 @@ var ItemSetLieutenantCommandersProwess = core.NewItemSet(core.ItemSet{
...
+			c := agent.GetCharacter()
+			c.AddStats(stats.Stats{
+				stats.Agility:           20,
+				stats.Spellpower:        23,
+			})
+		},
+	},
+})
+
+var ItemSetWarlordsProwess = core.NewItemSet(core.ItemSet{
```
kayla-glick, line 162: The Prowess sets also have a +4% raptor strike damage mod on the gloves if you wouldn't mind adding those in `sim/hunter/items.go` too
image

kayla-glick (commented):
Thanks for contributing! Just a couple things but always appreciate getting new folks involved!

sim/hunter/item_sets_pvp.go
Kawney: I believe both this pair and the other pair are included in this pull from Docoda:
https://github.com/wowsims/sod/pull/1001

kayla-glick: That's totally fine then! I know @ncberman is taking a look at Docoda's PR to get it cleaned up and passing so we'll get them there!

kayla-glick (approved):

PR #1028, opened 2024-09-06
Prot Paladin: Shield of Righteousness

wsphillips:
Changes:

* Added Shield of Righteousness rune ability
* Removed the rune from unimplemented overrides
* Scraped rune tooltips and regenerated db
* Updated default paladin APL to include Shield of Righteousness and updated tests.


sim/paladin/shield_of_righteousness.go
```diff
@@ -0,0 +1,49 @@
...
+func (paladin *Paladin) registerShieldOfRighteousness() {
+	if !paladin.hasRune(proto.PaladinRune_RuneCloakShieldOfRighteousness) {
+		return
+	}
+
+	level := float64(paladin.Level)
+
+	// Base damage formula from wowhead tooltip:
+	// https://www.wowhead.com/classic/spell=440658/shield-of-righteousness
+	damage := 179.0 * (9.046514 + 0.676562*level + 0.019349*level*level) / 100.0
```
kayla-glick, line 18: Every class has a function `baseRuneAbilityDamage()` that you can use to get this

wsphillips: Added baseRuneAbilityDamage and used it here.

PR #1030, opened 2024-09-06
Prot Paladin: Reckoning

wsphillips:
Implements the base Reckoning talent. The behavior should be correct while continuously autoattacking.

_Technically_ if you stop autoattacking, Reckoning can continue to proc, storing extra attacks up to a limit of 4 saved extra attacks. When you resume autoattacking, you swing 5 attacks all at once (1 regular swing + 4 extra attacks).

While we have the ability to store autoattacks now, I'm not sure how you would add this niche behavior, or whether its worth doing/useful...

To get the behavior described above you would need:

1. Probably a bespoke "stopattack" macro--You would want to stop attacking immediately after an autoswing and not "start" attacking again until your swing timer is off CD again.
2. A counter for the number of Reckoning procs accumulated while not attacking. (i.e. stop calling ExtraMHAttack after 4 procs)
3. A callback that resets the Reckoning counter once the extra attacks go out.

Curious to know what @Adamrch thinks.


Adamrch:
1) You can start auto attacking again right after you get the reckoning proc, it won't trigger until your next auto
2) you can keep calling StoreExtraMHAttack without worrying about storing more than 4, it handles that. ExtraMHAttack uses them right away and doesn't have a limit.
3) Why not just use the existing storedExtra attacks?

Adamrch:
I just tested it, you can start auto attacking before your swing timer is ready, As longer as you are not auto attacking when it procs.

wsphillips:
Oh wow....so:
1. I didn't see that you can manually call StoreExtraMHAttack.
2. I did not know 4 extra attacks was a universal limit, and that you already have it capped like that. I thought it was just something they did specifically for Paladins Reckoning after the infamous [Kazzak one-shot incident.](https://www.youtube.com/watch?v=7_Qb4P4yb9E)

That makes "full" implementation potentially way easier...let me dig into this a little more. I just need to get a better understanding of what you did with these stop attack macro flags

wsphillips:
Going to punt extensions for stopattack/storing attacks until we have tests showing it can work in practice and that it results in meaningful gains.

wsphillips:
@kayla-glick can has merge? Aegis rune + T2 depend on it.

kayla-glick (approved):

PR #1041, opened 2024-09-08
Warlock T2 tank set bonus

Shaklor:
Implemented Mongoose Bite Debuff to apply Dodge debuff as a sim option. This doesnt appear to be effecting DPS.
Warlock T2 bonuses not currently working.


sim/warlock/item_sets_pve.go
```diff
@@ -284,3 +284,75 @@ var ItemSetWickedFelheart = core.NewItemSet(core.ItemSet{
...
+			warlock.RegisterAura(core.Aura{
+				Label:    "S03 - Item - T2 - Warlock - Tank 4P Bonus",
+				Duration: core.NeverExpires,
+				OnReset: func(aura *core.Aura, sim *core.Simulation) {
+					aura.Activate(sim)
+				},
+				OnSpellHitDealt: func(aura *core.Aura, sim *core.Simulation, spell *core.Spell, result *core.SpellResult) {
+					if spell.SpellCode == SpellCode_WarlockShadowburn && result.Landed() {
+						healAmount := warlock.MaxHealth() * 0.15
+						warlock.GainHealth(sim, healAmount, warlock.NewHealthMetrics(core.ActionID{SpellID: 18871}))
```
kayla-glick, line 322: You should initialize the health metrics outside of the aura so that it doesn't have to create them every time a shadowburn hits

sim/warlock/item_sets_pve.go
```diff
@@ -284,3 +284,75 @@ var ItemSetWickedFelheart = core.NewItemSet(core.ItemSet{
...
+					}
+				},
+			})
+		},
+		// The 6-piece bonus creates a shield from excess healing, up to 30% of the Warlock's maximum health.
+		6: func(agent core.Agent) {
+			warlock := agent.(WarlockAgent).GetWarlock()
+
+			maxShieldAmount := warlock.MaxHealth() * 0.3
```
kayla-glick, line 331: Is the shield Dynamic? If the lock gained a +stam effect in combat I would assume the shield could be higher?

PR #1048, opened 2024-09-10
Prot Paladin: Malleable Protection rune + Divine Protection + Forbearance

wsphillips:
* Malleable Protection implemented.
* Divine Protection implemented (as modified by Malleable Protection)
* Forbearance debuff added

MP tooltip text for reference:

> You can now take actions during Divine Protection and its duration is increased by 50%, but it only reduces damage you take by 50% instead of making you immune. In addition, activating Holy Shield also grants you 4 Attack Power for each point of your defense skill beyond 300, lasting 15 sec.


sim/paladin/forbearance.go
```diff
@@ -0,0 +1,33 @@
...
+			oldEffect := spell.ApplyEffects
+
+			spell.ApplyEffects = func(sim *core.Simulation, unit *core.Unit, spell *core.Spell) {
+				oldEffect(sim, unit, spell)
+				forbearanceAura.Activate(sim)
+			}
+
+			spell.ExtraCastCondition = func(sim *core.Simulation, target *core.Unit) bool {
+				return !forbearanceAura.IsActive()
+			}
```
kayla-glick, line 30: You may want to grab the old function as well just in case so that this doesn't accidentally obscure adding extra cast conditions in forbearance spells and them not working correctly

kayla-glick (approved):

PR #1050, opened 2024-09-11
Hunter Trap Spellhit fix + Trap Talent/Rune Refactoring

ncberman:
Used a small workaround to have traps be agnostic of spellhit stats, school buffs, and debuffs
refactored TNT and Resourcefulness runes to apply modifiers on registration rather than in the spell code
refactored Trap Mastery and Clever Traps talents to apply modifiers on registration rather than in the spell code


ncberman:
Ptr update says that ring runes should impact trap hit chance now, since ring runes modify school hit chance I updated the code to no longer remove the school hit bonus when calculating trap hit chance.

sim/hunter/runes.go
```diff
@@ -364,18 +371,18 @@ func (hunter *Hunter) trapRange() float64 {
...
-	if hunter.HasRune(proto.HunterRune_RuneCloakResourcefulness) {
-		return 0.6
-	}
-	return 1.0
+	hunter.OnSpellRegistered(func(spell *core.Spell) {
+		if spell.Flags.Matches(SpellFlagTrap) {
+			// Ideally I would set Multiplier to 0 but that results in the cost calculation defaulting to 100
+			spell.Cost.FlatModifier = int32(spell.Cost.BaseCost * -1)
+			spell.CD.Duration = spell.CD.Duration / 100 * 60
+		}
```
kayla-glick, line 384: Couldn't you just change `spell.Cost.BaseCost` to 0?

ncberman: Oh, I guess idk I thought base cost was different than  flat cost but now I'm wondering if I'm confusing 2 different structs

kayla-glick: They are. Flat cost is a raw mana cost (e.g. 300) and Base cost is a % of total mana. It looks like traps are a flat cost so actually youd want to do `spell.Cost.FlatCost = 0` if you can

ncberman: I think that's where I was confused, I'm not sure FlatCost exists outside of the spell configuration SpellConfig.ManaCost struct. Seems like setting the Cost.BaseCost of the spell post construction is the correct way to set it.

kayla-glick: Oh I see. Ok that's fine then

PR #1052, opened 2024-09-11
Wing clip no longer consumes Clever Strikes

ncberman:
image
Most of this stuff has already been implemented or are bug fixes. Everything in the sim already works the way these patch notes are written except for the wing clip change (in this PR) and the fire frost specialization runes which I will update the hunter trap PR with

### DIFF
--- a/sim/hunter/hunter.go
+++ b/sim/hunter/hunter.go
@@ -44,6 +44,7 @@ const (
 	SpellCode_HunterCarve
 	SpellCode_HunterCarveHit
 	SpellCode_HunterMongooseBite
+	SpellCode_HunterWingClip
 	SpellCode_HunterVolley
 
 	SpellCode_HunterPetFlankingStrike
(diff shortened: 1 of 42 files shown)


sim/hunter/hunter.go
```diff
@@ -44,6 +44,7 @@ const (
 	SpellCode_HunterCarve
 	SpellCode_HunterCarveHit
 	SpellCode_HunterMongooseBite
+	SpellCode_HunterWingClip
```
kayla-glick, line 47: I think you need to apply this to the wing clip spell itself still?

kayla-glick (approved):

kayla-glick (changes requested):
Wing clip will still consume the aura because of the `OnCastComplete`. I mentioned on Discord, but consider setting an `affectedSpells` array up in an `OnInit` and use that in the callbacks

kayla-glick (approved):

ncberman:
I'm going to double check functionality really quick, not sure there are any tests that use this set bonus

kayla-glick (approved):

PR #1053, opened 2024-09-12
Shockadin PTR updates

mlizard32:
Implemented Libram of Wrath https://www.wowhead.com/classic/item=232420/libram-of-wrath

**Shockadin Updates for PTR changes 9/11**

- Libram of Wrath changed to only gain stacks when Holy Shock deals damage. No longer stacks from healing. 
- Libram of Wrath damage per stack changed to 20% (was 30%). 
- Tier 2 Holy Paladin 2-piece critical strike chance of Holy Shock reduced to 5% (was 20%).
- Infusion of Light: This rune now reduces the remaining cooldown on Holy Shock by 3 seconds on a critical strike (was: resetting it completely).

**Additional changes**

- Exorcism set to no longer have partial resists as shown in Logs (super weird to me) 
- Holy Wrath correctly reduces cooldown if player has Purifying Power Rune 
- Updated gear set and APL for Libram of Wrath


sim/paladin/items.go
```diff
@@ -227,3 +234,51 @@ func crusadersZealAura465414(character *core.Character) *core.Aura {
 		},
 	})
 }
+
+func libramOfWrathAura(agent core.Agent) {
```
kayla-glick, line 238: Is this shared with something else? If not it doesn't need to be a separate function

sim/paladin/items.go
```diff
@@ -227,3 +234,51 @@ func crusadersZealAura465414(character *core.Character) *core.Aura {
...
+
+	paladin.RegisterAura(core.Aura{
+		Label:    "Libram Of Wrath Trigger",
+		Duration: core.NeverExpires,
+		OnReset: func(aura *core.Aura, sim *core.Simulation) {
+			
+			aura.Activate(sim)
+		},
+		OnSpellHitDealt: func(aura *core.Aura, sim *core.Simulation, spell *core.Spell, result *core.SpellResult) {
+			if spell.SpellCode == SpellCode_PaladinHolyShock {
```
kayla-glick, line 278: Does the shock need to hit to proc the libram or can a full resist still proc?

sim/paladin/exorcism.go
```diff
@@ -51,7 +51,7 @@ func (paladin *Paladin) registerExorcism() {
 			SpellSchool: core.SpellSchoolHoly,
 			DefenseType: core.DefenseTypeMagic,
 			ProcMask:    core.ProcMaskSpellDamage,
-			Flags:       core.SpellFlagMeleeMetrics | core.SpellFlagAPL,
+			Flags:       core.SpellFlagMeleeMetrics | core.SpellFlagAPL | core.SpellFlagIgnoreResists,//Logs show it never has partial resists, No clue why, still misses
```
kayla-glick, line 54: I think the flag you're looking for here is `SpellFlagBinary`. `SpellFlagIgnoreResists` causes it to not be affected by resistance stats

kayla-glick (changes requested):
Just some minor things

sim/paladin/items.go
mlizard32: went and tested, it does proc on full resist - I changed it to OnCastComplete is that the appropriate trigger with that?

sim/paladin/exorcism.go
mlizard32: Well there is no Holy Resistance in the game, and IDK what is Binary about the spell, maybe it is flagged as binary? IDK how to look that up. Results are the same so I updated it

sim/paladin/items.go
kayla-glick: Then yeah it's probably an on cast complete!

sim/paladin/exorcism.go
kayla-glick: Oh ok sorry @mlizard32 could you add that flag back per what @wsphillips said?

mlizard32: Ya it full resists, which I believe is a "Miss", with either flag it will "Miss". 
Its super weird, on the era logs you can see it has partial resists, its some weird qurik with the SoD version. 
The end result is the same with either flag, I'm not really sure which makes more sense

mlizard32: I'm kinda wondering if the "100% crit if demon/undead" somehow needed to be coded in a "Binary" way so thats why it behaves like a "Binary" spell now?

kayla-glick: Ok are we good then do you think? That was my only concern so I can merge this otherwise

mlizard32: K I think It is "Binary" and here is my reasoning.

Binary spells are spells that have additional effects when cast, examples I looked at (pyroblast, shadow flame, frostfire bolt) these spells do not seem to partially resist according to logs I looked at, while all spells I looked at (shadow bolt, Holy Shock, Fire blast) with 1 simple effect do partially resist.

Comparing original Exorcism to new Exorcism, the new Exorcism has an additional "Effect" (a dummyscript).
Since the new spell has an additional "Effect" it is now a binary spell.

https://www.wowhead.com/classic/spell=415073/exorcism
https://www.wowhead.com/classic/spell=10314/exorcism

My vote is to leave the flag as Binary - @kayla-glick had it right.

kayla-glick (approved):

PR #1062, opened 2024-09-14
Minor fixes for Ranged Attack Power bonuses and applications

ncberman:
Juju Might updated to provide 40rAP in addition to mAP. 
https://www.wowhead.com/classic/spell=16329/juju-might

Predator's Armor 2P now provides 20rAP in addition to mAP.
https://www.wowhead.com/classic/spell=9331/attack-power-20#used-by-item-set

Dragonstalker's Pursuit 6P bonus to serpent sting no longer includes hunter's mark and mob-type rAP bonuses into its calculation.
https://www.wowhead.com/classic/spell=467326/s03-item-t2-hunter-ranged-6p-bonus (not sure if wowhead describes this interaction but testing on ptr shows the benefit is calculated from player stats)

Chimera - Serpent portion of Chimera shot has to roll individually to hit as its own ranged attack, chimera shot landing does not guarantee Chimera - Serpent will hit.

### DIFF
--- a/sim/core/spell_result.go
+++ b/sim/core/spell_result.go
@@ -95,10 +95,12 @@ func (spell *Spell) MeleeAttackPower() float64 {
 	return spell.Unit.stats[stats.AttackPower] + spell.Unit.PseudoStats.MobTypeAttackPower
 }
 
-func (spell *Spell) RangedAttackPower(target *Unit) float64 {
-	return spell.Unit.stats[stats.RangedAttackPower] +
+func (spell *Spell) RangedAttackPower(target *Unit, ignorePseudoStats bool) float64 {
+	return TernaryFloat64(ignorePseudoStats, 
+		spell.Unit.stats[stats.RangedAttackPower], 
+		spell.Unit.stats[stats.RangedAttackPower] +
 		spell.Unit.PseudoStats.MobTypeAttackPower +
-		target.PseudoStats.BonusRangedAttackPowerTaken
+		target.PseudoStats.BonusRangedAttackPowerTaken)
 }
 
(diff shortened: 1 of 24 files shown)


sim/core/spell_result.go
```diff
@@ -95,10 +95,12 @@ func (spell *Spell) MeleeAttackPower() float64 {
 	return spell.Unit.stats[stats.AttackPower] + spell.Unit.PseudoStats.MobTypeAttackPower
 }
 
-func (spell *Spell) RangedAttackPower(target *Unit) float64 {
-	return spell.Unit.stats[stats.RangedAttackPower] +
+func (spell *Spell) RangedAttackPower(target *Unit, ignorePseudoStats bool) float64 {
```
kayla-glick, line 98: Knit: Could we call this "ignore target modifiers"?

sim/hunter/serpent_sting.go
```diff
@@ -52,7 +52,7 @@ func (hunter *Hunter) getSerpentStingConfig(rank int) core.SpellConfig {
 			BonusCoefficient: spellCoeff,
 
 			OnSnapshot: func(sim *core.Simulation, target *core.Unit, dot *core.Dot, isRollover bool) {
-				damage := baseDamage + (hunter.SerpentStingAPCoeff*dot.Spell.RangedAttackPower(target))/5
+				damage := baseDamage + (hunter.SerpentStingAPCoeff*dot.Spell.RangedAttackPower(target, true))/5
```
kayla-glick, line 55: Could you leave a comment on these for any future person to know? Just something like "Serpent sting's bonus RAP ignores target modifiers like mob type AP and Hunter's Mark"

kayla-glick (approved):

PR #1066, opened 2024-09-16
Prot Paladin: Libram of the Devoted + Libram of Avengement

wsphillips:
Implements the Phase 4/5 Paladin librams.


sim/paladin/items.go
```diff
@@ -238,6 +240,62 @@ func init() {
...
+					// Second hit
+					timeToHit := spell.TravelTime() + time.Millisecond*1500
+					core.StartDelayedAction(sim, core.DelayedActionOptions{
+						DoAt: sim.CurrentTime + timeToHit,
+						OnAction: func(s *core.Simulation) {
+							spell.DealDamage(sim, results[1])
+						},
+					})
+				}
+			}
```
kayla-glick, line 295: I wonder if this could just be a second cast of avenger's shield but then manually adjusting the cast metrics?

kayla-glick: Oh gotcha. Hm, I wonder if this should just be a toggle on the avenging wrath spell then with an attribute you can override here? I don't love the spell's code being copied here

kayla-glick: Yeah idk sometimes these effects are a total pain lol. You could just tie it to a paladin struct attribute and have the libram toggle that attribute on. Something like

`paladin.DoRighteousShieldSingleTarget` or something idk. In Wrath it was just a glyph IIRC

kayla-glick (approved):

PR #1074, opened 2024-09-19
Rogue Presets v2

sanguinerarogue:


ui/rogue/presets.ts
```diff
@@ -271,13 +346,14 @@ export const DefaultIndividualBuffs = IndividualBuffs.create({
 	spiritOfZandalar: true,
 	valorOfAzeroth: true,
 	warchiefsBlessing: true,
+	spiritOfZandalar: true,
```
kayla-glick, line 349: This got screwed up in the merging but this just needs to be removed. It's already set above

PR #1077, opened 2024-09-20
Bug fix: Defer calculation of Block Value's strength dependency and add Shaman SBV from strength

### LINKED ISSUE #1025: Shield Block Value Multiplier only applies to SBV gained through flat bonuses
Currently the SBV multiplier is applied to total SBV:

https://github.com/wowsims/sod/blob/2bef638dc1fc917163940ee7a1cd4578bfc416ad/sim/core/unit.go#L359-L361

However, from testing (with my paladin) the in-game SBV multiplier ignores the portion of SBV gained via strength. The final amount of SBV is approximately:

```
Total SBV  = [(Strength/20) - 1] + [(Equipped raw SBV + dynamic raw SBV bonuses) * SBV Multiplier]
```

The 1 point offset + rounding are small enough to ignore.

Warriors (afaik) do not currently have any talents/runes/items that modify the SBV multiplier. So this appears to be a Paladin- and Shaman-specific issue.

I initially tried solving this with `ApplyDynamicEquipScaling` but the issue persists because the SBV per Strength stat dependency is applied to equipment stats before the equip scaling multiplier.

To get this working right, I would need dynamic equipment scaling that ignores stat dependencies and includes activated bonuses (e.g. Broodmother's Brooch). I'm not sure what the best solution is here so I'm opening it as an issue to get feedback.

wsphillips:
Fixes #1025 by adding `PseudoStat.BlockValuePerStrength` so that we can apply the `BlockValueMultiplier` independently to raw SBV and not SBV gained via strength.

Shield Block Value should now be correctly computed for all classes, including UI display.

### DIFF
--- a/sim/warrior/warrior.go
+++ b/sim/warrior/warrior.go
@@ -312,7 +312,7 @@ func NewWarrior(character *core.Character, talents string, inputs WarriorInputs)
 	warrior.PseudoStats.CanParry = true
 
 	warrior.AddStatDependency(stats.Strength, stats.AttackPower, core.APPerStrength[character.Class])
-	warrior.AddStatDependency(stats.Strength, stats.BlockValue, .05) // 20 str = 1 block
+	warrior.PseudoStats.BlockValuePerStrength = .05 // 20 str = 1 block
 	warrior.AddStatDependency(stats.Agility, stats.MeleeCrit, core.CritPerAgiAtLevel[character.Class][int(warrior.Level)]*core.CritRatingPerCritChance)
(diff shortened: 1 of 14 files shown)


sim/warrior/warrior.go
```diff
@@ -312,7 +312,7 @@ func NewWarrior(character *core.Character, talents string, inputs WarriorInputs)
 	warrior.PseudoStats.CanParry = true
 
 	warrior.AddStatDependency(stats.Strength, stats.AttackPower, core.APPerStrength[character.Class])
-	warrior.AddStatDependency(stats.Strength, stats.BlockValue, .05) // 20 str = 1 block
+	warrior.PseudoStats.BlockValuePerStrength = .05 // 20 str = 1 block
```
kayla-glick, line 315: Do shamans also get this bonus?

kayla-glick (approved):

PR #1078, opened 2024-09-21
fix seals and procs

Adamrch:


sim/core/consumes.go
```diff
@@ -393,7 +393,7 @@ func DragonBreathChiliAura(character *Character) *Aura {
 		ActionID:    ActionID{SpellID: 15851},
 		SpellSchool: SpellSchoolFire,
 		DefenseType: DefenseTypeMagic,
-		ProcMask:    ProcMaskEmpty,
+		ProcMask:    ProcMaskSpellDamage,
```
kayla-glick, line 396: Shouldn't chili be `ProcMaskSpellDamageProc`?

sim/mage/ignite.go
```diff
@@ -45,7 +45,7 @@ func (mage *Mage) applyIgnite() {
 		ActionID:    core.ActionID{SpellID: 12654},
 		SpellSchool: core.SpellSchoolFire,
 		DefenseType: core.DefenseTypeMagic,
-		ProcMask:    core.ProcMaskProc,
+		ProcMask:    core.ProcMaskSpellProc,
```
kayla-glick, line 48: I think this one would be `ProcMaskSpellDamageProc` right?

sim/paladin/soc.go
```diff
@@ -98,8 +98,8 @@ func (paladin *Paladin) registerSealOfCommand() {
 			ActionID:    core.ActionID{SpellID: rank.proc.spellID},
 			SpellSchool: core.SpellSchoolHoly,
 			DefenseType: core.DefenseTypeMelee,
-			ProcMask:    core.ProcMaskMeleeMHSpecial | core.ProcMaskProc,
-			Flags:       core.SpellFlagMeleeMetrics | SpellFlag_RV, // RV Worked on PTR
+			ProcMask:    core.ProcMaskMeleeMHSpecial | core.ProcMaskMeleeProc | core.ProcMaskMeleeDamageProc,
```
kayla-glick, line 101: Does SoC need both the normal melee proc mask and the melee damage proc mask or could it work with just the damage one?

sim/shaman/windfury_weapon.go
```diff
@@ -40,7 +40,7 @@ func (shaman *Shaman) newWindfuryImbueSpell(isMH bool) *core.Spell {
 		ActionID:    actionID,
 		SpellSchool: core.SpellSchoolPhysical,
 		DefenseType: core.DefenseTypeMelee,
-		ProcMask:    procMask | core.ProcMaskWeaponProc,
+		ProcMask:    procMask,
```
kayla-glick, line 43: I think the windfury weapon trigger also needs to be changed to not proc from procs so that phantoms don't proc it

sim/rogue/runes.go
```diff
@@ -96,7 +96,7 @@ func (rogue *Rogue) applyFocusedAttacks() {
 				isFoKOH = true
 			}
 
-			if !spell.ProcMask.Matches(core.ProcMaskMeleeOrRanged|core.ProcMaskTriggerInstant) || !result.DidCrit() || isFoKOH {
+			if !spell.ProcMask.Matches(core.ProcMaskMeleeOrRanged|core.ProcMaskMeleeDamageProc) || !result.DidCrit() || isFoKOH {
```
kayla-glick, line 99: Tested locally and focused attacks with blazefury seems to still be working correctly so that's awesome 🥂

sim/mage/ignite.go
Adamrch: I don't think so, because this is the application of the debuff,  the damage portion is taken care of in spell.dealDamageInternal
image

sim/rogue/runes.go
Adamrch: Nice!

sim/core/consumes.go
Adamrch: changed to ProcMaskSpellDamageProc | ProcMaskSpellProc so it procs both procs that require damage and those that don't.

sim/paladin/soc.go
Adamrch: it doesn't need it currently, but I'm implementing it this way so that it procs things that require damage in a melee, as well as those that only require it to be a melee proc.  I think a damaging melee proc should be flagged as a damage melee proc and a melee proc.

sim/shaman/windfury_weapon.go
Adamrch: added suppress equip flag

sim/mage/ignite.go
kayla-glick: Oh gotcha, ok

PR #1096, opened 2024-09-25
Rogue Level 60 Poison Implementation

sanguinerarogue:
Needs review, I believe the tags on the occult poison for Deadly Brew/Normal/Shiv are causing it to interfere and have different versions in sim if I use a manual imbue + deadly poison imbue.  I think this also is causing it to stack with the regular Occult Debuff from the settings picker.

image
image


sim/rogue/envenom.go
```diff
@@ -13,7 +13,7 @@ func (rogue *Rogue) registerEnvenom() {
 	}
 
 	baseAbilityDamage := rogue.baseRuneAbilityDamage()
-
+	consumed := int32(0)
```
kayla-glick, line 16: We should set this at the top of the `ApplyEffects` just to be sure that there's no leakage across iterations

sim/rogue/poisoned_knife.go
```diff
@@ -15,6 +15,7 @@ func (rogue *Rogue) registerPoisonedKnife() {
 	poisonedKnifeMetrics := rogue.NewEnergyMetrics(core.ActionID{SpellID: 425012})
 	hasDeadlyBrew := rogue.HasRune(proto.RogueRune_RuneDeadlyBrew)
 	hasJustAFleshWound := rogue.HasRune(proto.RogueRune_RuneJustAFleshWound)
+	numStacks := 0.0
```
kayla-glick, line 18: Same as above here we should set this inside the `ApplyEffects`

PR #1101, opened 2024-09-26
ws wf rework

Adamrch:
Rework WS to work like https://github.com/magey/classic-warrior/wiki/Windfury-Totem#triggered-by-melee-spell
Autos have a cast batch where damage and result are calculated, and a damage batch where autos are applied.

This should allow two Instant stike GCD's in a row to proc WS with the second ability benefiting from the AP of the first proc.
(Same for autos but harder to time / align)

Refactored WS/ WF totem to share the same helper function code.

Still need to prevent WS/WF AP from turning off in the middle of a batch though.


Adamrch:
image
image

kayla-glick (approved):

PR #1102, opened 2024-09-26
[UI] Migrate builds to preset configurations

1337LutZ:
- Migrated the Preset Configurations feature from Cata in favor of Preset builds. These presets will be shown on each tab that affects the preset.
- Things you can adjust in a configuration preset:
  - Gear
  - Talents
  - Rotation
  - EP Weights (new)
  - Encounter (new)
    - Encounter (boss/targets)
    - Healing Model
    - Tanks
    - Buffs
    - Debuffs
    - Raid buffs
    - Consumes
- Added option to show a toast warning if a user changes a setting but incompatible data is detected. Example can be found here: https://github.com/wowsims/cata/blob/master/ui/warrior/fury/presets.ts#L21


ui/tank_rogue/presets.ts
```diff
@@ -185,6 +185,23 @@ export const DefaultTalentsSubtlety = TalentPresets[Phase.Phase5][0];
 
 export const DefaultTalents = DefaultTalentsCombat;
 
+///////////////////////////////////////////////////////////////////////////
+//                                Encounters
+///////////////////////////////////////////////////////////////////////////
+export const PresetBuildEncounterDefault = PresetUtils.makePresetBuild('Default', {
```
1337LutZ, line 191: @sanguinerarogue I added these as an example, please feel free to modify these in this branch!

kayla-glick (approved):

PR #1109, opened 2024-09-29
Some sim reset fixes again

FelixPflaum:
[Core] Fix extraAttacksPending not being reset to 0 on sim reset.
[Shaman] Fix Power Surge rune aura incorrectly applying a dynamic stat dep OnInit.
[Mage] Fix Frozen Orb not resetting tick count on sim reset.
[Mage] Fix Arcanist Insight 6pc set bonus using OnInit to apply aura effects.
[Paladin] Fix Radiant Judgement 6pc bonus creating an aura dynamically.
[Warrior] Fix Execute rage resource event counter being off when spell cost is 0.

### DIFF
--- a/sim/mage/item_sets_pve.go
+++ b/sim/mage/item_sets_pve.go
@@ -150,13 +150,22 @@ var ItemSetArcanistInsight = core.NewItemSet(core.ItemSet{
 			mage.RegisterAura(core.Aura{
 				ActionID: core.ActionID{SpellID: 456402},
 				Label:    "S03 - Item - T1 - Mage - Damage 6P Bonus",
-				OnInit: func(aura *core.Aura, sim *core.Simulation) {
+				OnGain: func(aura *core.Aura, sim *core.Simulation) {
 					if mage.IceArmorAura != nil {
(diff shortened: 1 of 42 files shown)


sim/mage/item_sets_pve.go
```diff
@@ -150,7 +150,7 @@ var ItemSetArcanistInsight = core.NewItemSet(core.ItemSet{
 			mage.RegisterAura(core.Aura{
 				ActionID: core.ActionID{SpellID: 456402},
 				Label:    "S03 - Item - T1 - Mage - Damage 6P Bonus",
-				OnInit: func(aura *core.Aura, sim *core.Simulation) {
+				OnGain: func(aura *core.Aura, sim *core.Simulation) {
```
kayla-glick, line 153: Just curious why this can't be done in the OnInit? The FoF proc chance doesn't change dynamically during an encounter

FelixPflaum: Because OnInit only happens _once_ before the first iteration, so doing it like that the bonus is only active for the first iteration. Finally got to setup a dev VM for this again and want to get the concurrency port merged, and all of those showed up in the test there :D

PR #1118, opened 2024-10-08
implement off GCD apl actions

Adamrch:
image


Adamrch:
image

NerdEgghead:
At a high level, the way the Cata sim approaches this issue in a more general way is by attaching a separate Timer pointer to the Unit in addition to the existing GCD timer. The GCD timer is only used for `CanCast()` checks when appropriate, while the separate rotation timer is the one checked within APL. Dynamic events that can trigger rotational decisions, such as auto-attacks, must simply call a `ReactToEvent()` helper that automatically sets the rotation timer to be ready after a latency / reaction time delay, analogous to the 0.2s you have hardcoded in the posted example. Likewise, an APL can schedule an off-GCD rotational evaluation at a desired time in the future simply by executing the `Wait` or `WaitUntil` actions.

Unfortunately the current Cata rotation timing / spell queueing functionality was implemented piecemeal, so it's not neatly packaged into a single PR. I've gone through my commit history and put together a dump of all the relevant commits in order:

https://github.com/wowsims/cata/pull/141/commits/eb38dcdfc18f36a5d2b6eb8d83fdfafc7a480f3f
https://github.com/wowsims/cata/pull/141/commits/bcf38ffa34ed2ad461742a768d36a244cb09572b
https://github.com/wowsims/cata/pull/141/commits/4c5bf0053a2acbd1b1d069a474dbac7575f3fde8
https://github.com/wowsims/cata/pull/219/commits/46274e3e2b20de8ddb7d03318e653efebcc6b1ab
https://github.com/wowsims/cata/pull/219/commits/8ddad723521e2ef935b24b5fc52cb56da7ca5521
https://github.com/wowsims/cata/pull/219/commits/26422bc322b78b5d0cbbc7a39fe2897b47dddb52

Adamrch:
I'm not trying to model spell queueing, just model in the APL doing something after a delay. It shouldn't be a single static value. It should be something exposed to the API.  It seems to be working just fine without modeling spellqueueing. 
It is also intentional to be able to set the value to what I want as I cant press keys consecutively in game at any timing I want. 50ms (two keys next to each other), 100 ms, (mouse wheel up followed by down) or just mashing fast.

Adamrch:
I'm not interested in porting spellqueueing over from cata at the moment because it is something completely different than what I am trying to accomplish, and would not serve the purpose I need for the ret rotation.

Adamrch:
I believe if we are so focused on keeping the APL the same across more versions of the game maybe we should have a git submodule for APL (that's shared across SoD and Cata)

Adamrch:
"Dynamic events that can trigger rotational decisions, such as auto-attacks, must simply call a ReactToEvent()"   does this let you set the reaction time when you create the APL action? If I had two such actions could I set a different time for each?

Issue #1146, opened 2024-11-15
Phase 6 - Expertise


kayla-glick:
Phase 6 adds a new "expertise" stat that provides -1% chance to be dodged or parried. We still have an Expertise stat in the sim, however it's not currently used in the attack table calculations. We need to add it back in, then allow it to be parsed via Wowhead tooltips

PR #1148, opened 2024-11-15
Add expertise stat

FelixPflaum:
Reuse the existing expertise stat, treating it as simple percentages like hit and crit.
Update expertise regex to look for "Equip: Reduces the chance for your attacks to be dodged or parried by x%." on items.
Add expertise to shown stats and EP stats for all melee specs.
Also adds Symbols of Unending Life expertise set bonus for druid.


kayla-glick:
Resolves https://github.com/wowsims/sod/issues/1146

kayla-glick (approved):

Issue #1139, opened 2024-11-14
Phase 6 - Timeworn Effect


kayla-glick:
Phase 6 adds new bonuses for using non-tier raid items through the Brood of Nozdormu rings. There are 7 unique effects that give bonuses based on the number of "Timeworn" non-tier raid AQ raid pieces. There are also 5 variants of each ring (Neutral, Friendly, Honored, Revered, Exalted) that need to be accounted for.

These are the 7 unique rings:
- [x] https://www.wowhead.com/classic/item=234440/signet-ring-of-the-bronze-dragonflight
- [x] https://www.wowhead.com/classic/item=234968/signet-ring-of-the-bronze-dragonflight
- [x] https://www.wowhead.com/classic/item=234198/signet-ring-of-the-bronze-dragonflight
- [x] https://www.wowhead.com/classic/item=234019/signet-ring-of-the-bronze-dragonflight
- [x] https://www.wowhead.com/classic/item=234018/signet-ring-of-the-bronze-dragonflight
- [x] https://www.wowhead.com/classic/item=234017/signet-ring-of-the-bronze-dragonflight
- [x] https://www.wowhead.com/classic/item=234016/signet-ring-of-the-bronze-dragonflight

Each has 5 items that need to be accounted for.

PR #1154, opened 2024-11-16
add conqueror protector and dominator timeworn effects

Adamrch:


sim/core/stats/stats.go
```diff
@@ -62,6 +62,7 @@ const (
 	HealingPower
 	SpellDamage
 	FeralAttackPower
+	Timeworn
```
FelixPflaum, line 65: Don't add a stat for this. See the note below that explicitly says not to add stats :D 

If at all make a pseudo stat for it, or just a stand-alone function that counts valid equipped items and use that inside the aura handlers for the rings.

sim/common/sod/item_effects/phase_6.go
```diff
@@ -0,0 +1,122 @@
...
+	SignetRingSubjugatorR5  = 234436
+	SignetRingSubjugatorR4  = 234437
+	SignetRingSubjugatorR3  = 234438
+	SignetRingSubjugatorR2  = 234439
+	SignetRingSubjugatorR1  = 234440
+	SignetRingConquerorR5   = 234202
+	SignetRingConquerorR4   = 234201
+	SignetRingConquerorR3   = 234200
+	SignetRingConquerorR2   = 234199
+	SignetRingConquerorR1   = 234198
```
kayla-glick, line 45: Normally I like to keep these ordered by ID but I think this makes a lot of sense like this here 👍 keeps it much more readable

kayla-glick:
Resolves https://github.com/wowsims/sod/issues/1139

sim/core/stats/stats.go
kayla-glick: Converted to a pseudostat

PR #1156, opened 2024-11-16
Rogue Tank AQ Set

sanguinerarogue:


sim/rogue/items_sets_pve.go
```diff
@@ -440,4 +440,108 @@ var ItemSetDeathdealersThrill = core.NewItemSet(core.ItemSet{
...
\ No newline at end of file
+})
+
+var ItemSetDeathdealersBattlearmor = core.NewItemSet(core.ItemSet{
+	Name: "Deathdealer's Battlearmor",
+	Bonuses: map[int32]core.ApplyEffect{
+		// Your Main Gauche now strikes 1 additional nearby target and also causes your Sinister Strike to strike 1 additional nearby target. 
+		// These additional strikes are not duplicated by Blade Flurry.
+		2: func(agent core.Agent) {
+			//Implemented in sinister_strike.go and main_gauche.go
```
kayla-glick, line 451: I think this comment is a copy-paste? 😂

sim/rogue/items_sets_pve.go
```diff
@@ -440,4 +440,108 @@ var ItemSetDeathdealersThrill = core.NewItemSet(core.ItemSet{
...
+				ApplyEffects: func(sim *core.Simulation, target *core.Unit, spell *core.Spell) {
+					spell.CalcAndDealDamage(sim, target, curDmg, spell.OutcomeAlwaysHit)
+				},
+			})	
+			
+			cleaveAura := rogue.RegisterAura(core.Aura{
+				Label:    "2P Cleave Buff",
+				Duration: time.Second * 10,
+				OnSpellHitDealt: func(aura *core.Aura, sim *core.Simulation, spell *core.Spell, result *core.SpellResult) {
+					if rogue.Env.GetNumTargets() > 1 && result.Landed() && (spell.SpellCode == SpellCode_RogueSinisterStrike) {
```
kayla-glick, line 477: Instead of checking target counts within these callbacks, we know that this set is functionally useless on 1 target, so can we just not build all of these auras if `rogue.Env.GetNumTarget() == 1`? Basically just adding it to the guard at the top of the set bonus. That should be a nice performance improvement

sim/rogue/items_sets_pve.go
```diff
@@ -440,4 +440,108 @@ var ItemSetDeathdealersThrill = core.NewItemSet(core.ItemSet{
...
+						curDmg = result.Damage / result.ResistanceMultiplier
+						cleaveHit.Cast(sim, rogue.Env.NextTargetUnit(result.Target))
+						cleaveHit.SpellMetrics[result.Target.UnitIndex].Casts--
```
kayla-glick, line 480: You could call `cleaveHit.ApplyEffects` here instead to avoid having to decrement metrics I think

sim/rogue/items_sets_pve.go
```diff
@@ -440,4 +440,108 @@ var ItemSetDeathdealersThrill = core.NewItemSet(core.ItemSet{
...
+				return
+			}
+			
+			var curDmg float64
+
+			cleaveHit := rogue.RegisterSpell(core.SpellConfig{
+				ActionID:    core.ActionID{SpellID: 1213754},
+				SpellSchool: core.SpellSchoolPhysical,
+				ProcMask:    core.ProcMaskEmpty,
+				Flags:       core.SpellFlagMeleeMetrics | core.SpellFlagNoOnCastComplete,
```
kayla-glick, line 463: We probably want to add the passive spell flag here as well

sim/rogue/talents.go
```diff
@@ -313,17 +313,18 @@ func (rogue *Rogue) registerBladeFlurryCD() {
 			rogue.MultiplyMeleeSpeed(sim, 1/1.2)
 		},
 		OnSpellHitDealt: func(aura *core.Aura, sim *core.Simulation, spell *core.Spell, result *core.SpellResult) {
-			isFoKOH := false
+			bfEligible := true
 
-			if spell.ActionID.SpellID == 409240 && spell.ActionID.Tag == 2 {
-				isFoKOH = true
+			//Checks for FoK Offhand and 2P TAQ Set Piece Extra Hits.
+			if (spell.ActionID.SpellID == 409240 && spell.ActionID.Tag == 2) || spell.ActionID.SpellID == 1213754 {
```
kayla-glick, line 319: Could you potentially replace the SpellID checks here with spell codes? I know that these were already here so if it's a pain then don't worry about it for now

sanguinerarogue: I dont think we have SpellCodes for FoK because its only checking the Action ID tag which is the offhand.  So will leave this one as is.

Issue #1160, opened 2024-11-17
Create a new stat for +Weapon Damage


kayla-glick:
A select few items had a `+ X Weapon Damage` effect in Vanilla. In SoD this number has drastically increased. Right now we track this as a Pseudostat, meaning each item must be implemented manually rather than being parsed from tooltips and stored in the database. This stat is functionally the `Physical` version of the various `School Power` stats, so we should consider just creating a proper stat for it and parsing it in the database.

- [ ] Create a new stat `PhysicalPower` for tracking this stat
- [ ] Parse the stat from wowhead tooltips like we do other school powers
- [ ] Remove manual implementations of the pseudostat and the pseudostat altogether

PR #1165, opened 2024-11-18
Bonus weapon dmg parse and pseudo stats in display stats

FelixPflaum:
Add tooltip parsing for bonus weapon damage, remove manual item effects for it.
Rename BonusDamage pseudo stat to BonusPhysicalDamage to be more clear about what it is.
Add support for showing pseudo stats in char stats UI, add timeworn count.
Add bonus weapon damage as possible stat for stat weights UI.
Add bonus weapon damage for item EP calculation.
Add bonus ewapon damage as feral display and EP stat.

Edit: No clue why rogue P2 test results are different with this, probably using an item with bonus dmg on it?

grafik
grafik


kayla-glick:
Resolves https://github.com/wowsims/sod/issues/1160

kayla-glick (approved):

PR #1185, opened 2024-11-28
Implementing the bear sim

emsimpson92:


ui/core/proto_utils/names.ts
```diff
@@ -207,7 +207,7 @@ export const pseudoStatNames: Map<PseudoStat, string> = new Map([
 	[PseudoStat.PseudoStatOffHandDps, 'Off Hand DPS'],
 	[PseudoStat.PseudoStatRangedDps, 'Ranged DPS'],
 	[PseudoStat.PseudoStatBlockValueMultiplier, 'Block Value Multiplier'],
-	[PseudoStat.TimewornBonus, 'Timeworn Pieces'],
+	[PseudoStat.TimewornBonus, 'Timeworn Pieces2'],
```
kayla-glick, line 210: Is this an accidental change?

ui/core/components/sim_title_dropdown.tsx
```diff
@@ -168,7 +168,7 @@ export class SimTitleDropdown extends Component {
 				<div className="sim-link-content">
 					<img src={this.getSimIconPath(data)} className="sim-link-icon" />
 					<div className="d-flex flex-column">
-						<span className="sim-link-label text-white">WoWSims - Season of Discovery</span>
+						<span className="sim-link-label text-white">WoWSims - Season of Discovery2</span>
```
kayla-glick, line 171: Is this an accident?

ui/core/proto_utils/names.ts
emsimpson92: Yeah I'm not sure what made these changes but I'll revert them

sim/common/vanilla/item_effects.go
```diff
@@ -3296,3 +3315,31 @@ func manslayerOfTheQirajiAura(character *core.Character) *core.Aura {
...
+		ApplyEffects: func(sim *core.Simulation, target *core.Unit, spell *core.Spell) {
+			spell.CalcAndDealDamage(sim, target, damage, spell.OutcomeAlwaysHit)
+		},
+	})
+
+	core.MakeProcTriggerAura(&character.Unit, core.ProcTrigger{
+		Name:     procName,
+		Callback: core.CallbackOnSpellHitTaken,
+		Outcome:  core.OutcomeLanded,
+		ProcMask: core.ProcMaskMelee, // TODO: Unsure if this means melee attacks or all attacks
```
kayla-glick, line 3340: Re the comment, it means all melee attacks (white and yellow hits)

sim/druid/runes.go
```diff
@@ -25,7 +25,6 @@ func (druid *Druid) ApplyRunes() {
 	druid.applyElunesFires()
 
 	// Hands
-	druid.applyMangle()
```
kayla-glick, line 28: Removing this will make the sim not register the mangle cat spell I think. I would leave this here but remove the parts in `druid.go` that call `registerMangleBearSpell` since mangle is a rune in SOD

sim/druid/forms.go
```diff
@@ -263,144 +263,126 @@ func (druid *Druid) registerCatFormSpell() {
...
-// 				druid.AddRage(sim, rageDelta, rageMetrics)
-// 			} else if rageDelta < 0 {
-// 				druid.SpendRage(sim, -rageDelta, rageMetrics)
-// 			}
-// 			druid.BearFormAura.Activate(sim)
-// 		},
-// 	})
-// }
+func (druid *Druid) registerBearFormSpell() {
+	actionID := core.ActionID{SpellID: 9634}
```
kayla-glick, line 267: Since dire bear form is only available at level 40+ could you do the ID conditionally based on the bear's level?

sim/druid/forms.go
```diff
@@ -263,144 +263,126 @@ func (druid *Druid) registerCatFormSpell() {
...
+
+	statBonus := druid.GetFormShiftStats().Add(stats.Stats{
+		stats.AttackPower: 3 * float64(druid.Level),
+		stats.Health:      1240,
+	})
```
kayla-glick, line 272: Could you also do the same here based on level?

sim/druid/forms.go
```diff
@@ -263,144 +263,126 @@ func (druid *Druid) registerCatFormSpell() {
...
+
+			druid.AutoAttacks.SetMH(clawWeapon)
+
+			druid.PseudoStats.ThreatMultiplier += .3 + .03*float64(druid.Talents.FeralInstinct)
+			druid.PseudoStats.DamageTakenMultiplier *= sotfdtm
+
+			predBonus = druid.GetDynamicPredStrikeStats()
+			druid.AddStatsDynamic(sim, predBonus)
+			druid.AddStatsDynamic(sim, statBonus)
+			druid.ApplyDynamicEquipScaling(sim, stats.Armor, 4.6)
```
kayla-glick, line 309: Here too please!

sim/druid/forms.go
```diff
@@ -263,144 +263,126 @@ func (druid *Druid) registerCatFormSpell() {
...
+			druid.form = Humanoid
+			druid.SetCurrentPowerBar(core.ManaBar)
+			druid.AutoAttacks.SetMH(druid.WeaponFromMainHand())
+
+			druid.PseudoStats.ThreatMultiplier -= .3 + .03*float64(druid.Talents.FeralInstinct)
+			druid.PseudoStats.DamageTakenMultiplier /= sotfdtm
+
+			druid.AddStatsDynamic(sim, predBonus.Invert())
+			druid.AddStatsDynamic(sim, statBonus.Invert())
+			druid.RemoveDynamicEquipScaling(sim, stats.Armor, 4.6)
```
kayla-glick, line 335: And last one here I think

sim/druid/frenzied_regeneration.go
```diff
@@ -7,7 +7,7 @@ import (
 )
 
 func (druid *Druid) registerFrenziedRegenerationCD() {
-	actionID := core.ActionID{SpellID: 22842}
+	actionID := core.ActionID{SpellID: 22896}
```
kayla-glick, line 10: Can we also do this talent by level? There seem to be 3 ranks for 10/15/20 rage accordingly
https://www.wowhead.com/classic/spell=22842/frenzied-regeneration
https://www.wowhead.com/classic/spell=22895/frenzied-regeneration
https://www.wowhead.com/classic/spell=22896/frenzied-regeneration

sim/druid/druid.go
```diff
@@ -228,20 +247,21 @@ func (druid *Druid) RegisterFeralCatSpells() {
...
-	// druid.registerRakeSpell()
-	// druid.registerRipSpell()
+	druid.registerEnrageSpell()
+	druid.registerFrenziedRegenerationCD()
+	druid.registerMangleBearSpell()
+	druid.registerRakeSpell()
+	druid.registerRipSpell()
+	druid.registerMaulSpell(realismICD)
+	druid.registerLacerateBleedSpell()
+	druid.registerLacerateSpell()
```
kayla-glick, line 262: Can we move the lacerate registration to the `runes.go` file for consistency?

sim/druid/lacerate.go
```diff
@@ -0,0 +1,121 @@
...
+
+	druid.Lacerate = druid.RegisterSpell(Bear, core.SpellConfig{
+		ActionID:    core.ActionID{SpellID: 414644},
+		SpellSchool: core.SpellSchoolPhysical,
+		DefenseType: core.DefenseTypeMelee,
+		ProcMask:    core.ProcMaskMeleeMHSpecial,
+		Flags:       SpellFlagOmen | core.SpellFlagMeleeMetrics | core.SpellFlagAPL,
+
+		RageCost: core.RageCostOptions{
+			Cost:   core.TernaryFloat64(druid.BerserkAura.IsActive(), 0, 10),
```
kayla-glick, line 30: This only runs during initialization, so you'll have to dynamically modify the cost inside the aura's effects

sim/druid/lacerate.go
```diff
@@ -0,0 +1,121 @@
...
+		},
+
+		DamageMultiplier: initialDamageMul,
+		ThreatMultiplier: 3.25,
+		// TODO: Berserk 3 target lacerate cleave - Saeyon
+
+		ApplyEffects: func(sim *core.Simulation, target *core.Unit, spell *core.Spell) {
+			baseDamage := (spell.Unit.MHWeaponDamage(sim, spell.MeleeAttackPower()) * .2) * float64(druid.LacerateBleed.Dot(target).GetStacks())
+
+			spell.DamageMultiplier = initialDamageMul
```
kayla-glick, line 47: I don't think this needs to be assigned here?

sim/druid/lacerate.go
```diff
@@ -0,0 +1,121 @@
...
+			result := spell.CalcAndDealDamage(sim, target, baseDamage, spell.OutcomeMeleeSpecialHitAndCrit)
```
kayla-glick, line 48: I looked at my guild's bear's log and it seems like lacerate can't miss? Maybe WCL just shows it wrong?
image

sim/druid/lacerate.go
```diff
@@ -0,0 +1,121 @@
...
+
+			if result.Landed() {
+				druid.LacerateBleed.Cast(sim, target)
+
+				if druid.HasRune(proto.DruidRune_RuneHelmGore) && sim.Proc(0.15, "Gore") {
```
kayla-glick, line 53: Rune checks are really expensive and can be cached as variables, so can perform the rune check at the top of the function and assign it to something like `hasGoreRune`?

sim/druid/lacerate.go
```diff
@@ -0,0 +1,121 @@
...
+			},
+			IgnoreHaste: true,
+		},
+
+		DamageMultiplier: initialDamageMul,
+		ThreatMultiplier: 3.25,
+		// TODO: Berserk 3 target lacerate cleave - Saeyon
+
+		ApplyEffects: func(sim *core.Simulation, target *core.Unit, spell *core.Spell) {
+			baseDamage := (spell.Unit.MHWeaponDamage(sim, spell.MeleeAttackPower()) * .2) * float64(druid.LacerateBleed.Dot(target).GetStacks())
```
kayla-glick, line 45: I could be wrong but on wowhead it looks like this ability uses the rune scaling formula https://www.wowhead.com/classic/spell=414644/lacerate

sim/druid/lacerate.go
```diff
@@ -0,0 +1,121 @@
...
+			TickLength:    time.Second * 3,
+
+			OnSnapshot: func(sim *core.Simulation, target *core.Unit, dot *core.Dot, isRollover bool) {
+				dot.SnapshotBaseDamage = tickDamage
+				dot.SnapshotBaseDamage *= float64(dot.Aura.GetStacks())
+
+				if !isRollover {
+					attackTable := dot.Spell.Unit.AttackTables[target.UnitIndex][dot.Spell.CastType]
+					dot.SnapshotAttackerMultiplier = dot.Spell.AttackerDamageMultiplier(attackTable)
+				}
```
kayla-glick, line 101: If I'm reading the spell description right isn't this where we should be using the rune damage formula plus the weapon scaling times the number of stacks?

sim/druid/lacerate.go
```diff
@@ -0,0 +1,121 @@
...
+			dot := spell.Dot(target)
+			if dot.IsActive() {
+				dot.Refresh(sim)
+				dot.AddStack(sim)
+				dot.TakeSnapshot(sim, true)
+			} else {
+				dot.Apply(sim)
+				dot.SetStacks(sim, 1)
+				dot.TakeSnapshot(sim, true)
+			}
```
kayla-glick, line 118: These are functionally the same since `Apply` with just call `Refresh` already if the aura is active

sim/druid/lacerate.go
```diff
@@ -0,0 +1,121 @@
...
+
+func (druid *Druid) registerLacerateBleedSpell() {
+	if !druid.HasRune(proto.DruidRune_RuneLegsLacerate) {
+		return
+	}
+	tickDamage := 20.0
+
+	switch druid.Ranged().ID {
+	case IdolOfCruelty:
+		tickDamage += 7.0
```
kayla-glick, line 72: I think @FelixPflaum or maybe Balor said that this idol bumps the 20% weapon damage to 27% and that it's not a flat damage like the tooltip suggests

sim/druid/mangle.go
```diff
@@ -35,22 +44,22 @@ func (druid *Druid) registerMangleBearSpell() {
...
 				Timer:    druid.NewTimer(),
-				Duration: time.Duration(float64(time.Second) * (6 - durReduction)),
+				Duration: time.Second * 6,
 			},
 		},
+		// TODO: Berserk 3 target mangle cleave - Saeyon
 
-		DamageMultiplier: (1 + 0.1*float64(druid.Talents.SavageFury)) * 1.15
-		ThreatMultiplier: core.TernaryFloat64(druid.HasSetBonus(ItemSetThunderheartHarness, 2), 1.15, 1),
+		DamageMultiplier: (1.6 + 0.1*float64(druid.Talents.SavageFury)) * baseMultiplier,
```
kayla-glick, line 52: Both savage fury and ursine power apply additive multipliers (they uses `Modifies Damage/Healing done` so you'd want to add them to the `DamageMultiplierAddititve` field

sim/druid/maul.go
```diff
@@ -0,0 +1,106 @@
...
+
+func (druid *Druid) registerMaulSpell(realismICD *core.Cooldown) {
+	flatBaseDamage := 128.0
+	rageCost := 15 - float64(druid.Talents.Ferocity)
+
+	switch druid.Ranged().ID {
+	case IdolOfBrutality:
+		rageCost -= 3
+	}
+	rageMetrics := druid.NewRageMetrics(core.ActionID{SpellID: 431446})
```
kayla-glick, line 16: We should also do the level-based values here

sim/druid/maul.go
```diff
@@ -0,0 +1,106 @@
...
+		DefenseType: core.DefenseTypeMelee,
+		ProcMask:    core.ProcMaskMeleeMHSpecial | core.ProcMaskMeleeMHAuto,
+		Flags:       SpellFlagOmen | core.SpellFlagMeleeMetrics | core.SpellFlagNoOnCastComplete,
+
+		RageCost: core.RageCostOptions{
+			Cost:   rageCost,
+			Refund: 0.8,
+		},
+
+		DamageMultiplier: 1 + .1*float64(druid.Talents.SavageFury),
```
kayla-glick, line 31: Same here with savage fury

sim/druid/lacerate.go
emsimpson92: I haven't dug too deep into the review yet, but I believe this was handled here
https://github.com/wowsims/sod/pull/1185#discussion_r1866785045

sim/druid/lacerate.go
FelixPflaum: The triggered spell always hits. Seems to be done the other way round here. In the game you cast the bleed spell and the direct damage is triggered.

b4ugh:
I wanted to share some thoughts about the feedback regarding level scaling and level-based values. While I absolutely see the value in addressing this, I wonder if it might make sense to prioritize getting this initial version out first and consider level scaling as a follow-up.

With a seasonal server and where we are in the game’s lifecycle (the endgame), there’s a lot of excitement—especially from the classic Druid sod-tank channel—about experimenting with an alpha version of the sim as they prepare for AQ launch. Getting this out sooner would make a big impact for those eager to dive in, and we could always iterate on nice-to-have features like level scaling in a future PR.

sim/druid/lacerate.go
emsimpson92: Yeah there was a weird issue where I was unable to get the direct hit to crit when triggering the bleed first, so I just made the initial hit trigger the bleed instead

sim/druid/lacerate.go
emsimpson92: image
This is addressed in the direct damage lacerate spell

sim/druid/maul.go
emsimpson92: I'm not sure level scaling applies here. Savage fury scaling should be fine I think

sim/druid/lacerate.go
emsimpson92: I have the DOT and the damage spell separated into 2 different skills

sim/druid/lacerate.go
emsimpson92: replacing this with the following stopped lacerate from stacking beyond 1. I've reverted this until further feedback is received
`dot.Apply(sim)
dot.AddStack(sim)
dot.TakeSnapshot(sim, true)`

sim/druid/lacerate.go
emsimpson92: I don't think this is correct. After making this change, the direct damage dropped down to about 190 at 5 stacks. I've reverted this change until Balor can review, but the numbers I'm getting now are consistent with what I'm seeing on WCL

emsimpson92: Following up, I used the direct damage formula provided by voraciousghost in the screenshot below. The damage seems to be scaling properly with idol of cruelty

sim/druid/talents.go
```diff
@@ -173,42 +157,33 @@ func (druid *Druid) applyNaturesGrace() {
...
+			aura.Activate(sim)
+		},
+		OnSpellHitDealt: func(aura *core.Aura, sim *core.Simulation, spell *core.Spell, result *core.SpellResult) {
+			if druid.InForm(Bear) {
+				if result.Outcome.Matches(core.OutcomeCrit) {
+					if sim.Proc(procChance, "Primal Fury") {
+						druid.AddRage(sim, 5, rageMetrics)
+					}
+				}
+			}
```
kayla-glick, line 183: Ideally it would be nice to be able to just enable/disable this aura when entering/exiting bear form to save on checking the form each time. You reassign the bear form `OnGain` and `OnExpire` functions and have them enable/disable this aura

sim/druid/talents.go
```diff
@@ -173,42 +157,33 @@ func (druid *Druid) applyNaturesGrace() {
...
+		OnReset: func(aura *core.Aura, sim *core.Simulation) {
+			aura.Activate(sim)
+		},
+		OnSpellHitDealt: func(aura *core.Aura, sim *core.Simulation, spell *core.Spell, result *core.SpellResult) {
+			if druid.InForm(Bear) {
+				if result.Outcome.Matches(core.OutcomeCrit) {
+					if sim.Proc(procChance, "Primal Fury") {
+						druid.AddRage(sim, 5, rageMetrics)
+					}
+				}
```
kayla-glick, line 182:
```suggestion
				if result.Outcome.Matches(core.OutcomeCrit) && sim.Proc(procChance, "Primal Fury") {
					druid.AddRage(sim, 5, rageMetrics)
				}
```

sim/druid/swipe.go
```diff
@@ -62,18 +50,31 @@ func (druid *Druid) registerSwipeBearSpell() {
...
+				spell.BonusCritRating -= dotBonusCrit
+
 				target = sim.Environment.NextTargetUnit(target)
 			}
 
 			for _, result := range results {
 				spell.DealDamage(sim, result)
 			}
+
+			if druid.HasRune(proto.DruidRune_RuneHelmGore) && sim.Proc(0.15, "Gore") {
```
kayla-glick, line 74: Can you cache the rune check as a variable outside of the spell definition for performance?

sim/druid/swipe.go
```diff
@@ -62,18 +50,31 @@ func (druid *Druid) registerSwipeBearSpell() {
...
 		},
 
-		DamageMultiplier: 1 + 0.1*float64(druid.Talents.SavageFury),
+		DamageMultiplier: baseMultiplier + 0.1*float64(druid.Talents.SavageFury),
 		ThreatMultiplier: SwipeThreatMultiplier,
 
 		ApplyEffects: func(sim *core.Simulation, target *core.Unit, spell *core.Spell) {
 			for idx := range results {
+				dotBonusCrit := 0.0
+				if druid.LacerateBleed.Dot(target).GetStacks() > 0 {
```
kayla-glick, line 59: Will this potentially error if you don't have the Lacerate rune equipped? You might need to check if they have the lacerate rune (another cached call)

sim/druid/swipe.go
```diff
@@ -39,18 +26,19 @@ func (druid *Druid) registerSwipeBearSpell() {
...
 	case IdolOfBrutality:
 		rageCost -= 3
+	case IdolOfUrsinPower:
+		baseMultiplier += .03
 	}
+	rageMetrics := druid.NewRageMetrics(core.ActionID{SpellID: 431446})
 
 	druid.SwipeBear = druid.RegisterSpell(Bear, core.SpellConfig{
-		ActionID:    core.ActionID{SpellID: spellID},
+		ActionID:    core.ActionID{SpellID: 9908},
```
kayla-glick, line 35: Looks like swipe has 5 ranks. Can you split out the damage / actionID by level here?

sim/druid/swipe.go
```diff
@@ -62,18 +50,31 @@ func (druid *Druid) registerSwipeBearSpell() {
 			IgnoreHaste: true,
 		},
 
-		DamageMultiplier: 1 + 0.1*float64(druid.Talents.SavageFury),
+		DamageMultiplier: baseMultiplier + 0.1*float64(druid.Talents.SavageFury),
```
kayla-glick, line 53: A base damage multiplier should be applied to the `BaseDamageMultiplierAdditive` field. It looks like Savage Fury is an additive full damage multiplier so it should technically be applied to `DamageMultiplierAddtitive` instead, too

sim/druid/maul.go
kayla-glick: Sorry it's been a while so missing some context 😅 but since Savage Fury is additive it should be added to `DamageMultiplierAdditive` instead

sim/druid/maul.go
```diff
@@ -0,0 +1,167 @@
...
+
+			spell.BonusCritRating += dotBonusCrit
+			result := spell.CalcAndDealDamage(sim, target, baseDamage, spell.OutcomeMeleeSpecialHitAndCrit)
+			spell.BonusCritRating -= dotBonusCrit
+
+			if !result.Landed() {
+				spell.IssueRefund(sim)
+			}
+
+			if druid.HasRune(proto.DruidRune_RuneHelmGore) && sim.Proc(0.15, "Gore") {
```
kayla-glick, line 112: We should cache the HasRune here as well for performance

sim/druid/mangle.go
```diff
@@ -35,22 +45,33 @@ func (druid *Druid) registerMangleBearSpell() {
...
 				Timer:    druid.NewTimer(),
-				Duration: time.Duration(float64(time.Second) * (6 - durReduction)),
+				Duration: time.Second * 6,
 			},
 		},
+		// TODO: Berserk 3 target mangle cleave - Saeyon
 
-		DamageMultiplier: (1 + 0.1*float64(druid.Talents.SavageFury)) * 1.15
-		ThreatMultiplier: core.TernaryFloat64(druid.HasSetBonus(ItemSetThunderheartHarness, 2), 1.15, 1),
+		DamageMultiplier: 1.6 * (1 + 0.1*float64(druid.Talents.SavageFury)) * idolMultiplier,
```
kayla-glick, line 53: These are mixing the additive and multiplicative multipliers so we should modify this a bit
```suggestion
		DamageMultiplier: 1.6 * idolMultiplier,
		DamageMultiplierAdditive: 1 + 0.1*float64(druid.Talents.SavageFury)
```

sim/druid/mangle.go
```diff
@@ -95,9 +115,9 @@ func (druid *Druid) registerMangleCatSpell() {
 			IgnoreHaste: true,
 		},
 
-		DamageMultiplier: (1 + 0.1*float64(druid.Talents.SavageFury)) * weaponMulti,
-		ThreatMultiplier: 1,
-		BonusCoefficient: 1,
+		DamageMultiplierAdditive: (1 + 0.1*float64(druid.Talents.SavageFury)) * weaponMulti,
```
kayla-glick, line 118: This should also be split into additive / not additive. The weapon multi is multiplicative but Savage fury addittive
```suggestion
    DamageMultiplier: 1 * weaponMulti,
		DamageMultiplierAdditive: 1 + 0.1*float64(druid.Talents.SavageFury),
```

sim/druid/enrage.go
```diff
@@ -11,30 +11,22 @@ func (druid *Druid) registerEnrageSpell() {
 	actionID := core.ActionID{SpellID: 5229}
 	rageMetrics := druid.NewRageMetrics(actionID)
 
-	instantRage := []float64{20, 24, 27, 30}[druid.Talents.Intensity]
-
-	dmgBonus := 0.05 * float64(druid.Talents.KingOfTheJungle)
-
-	t10_4p := druid.HasSetBonus(ItemSetLasherweaveBattlegear, 4)
+	instantRage := []float64{0, 5, 10}[druid.Talents.ImprovedEnrage]
```
balor89, line 14: The [SoD tuning aura](https://www.wowhead.com/classic/spell=436895/s03-tuning-and-overrides-passive-druid) for druid makes a few changes to Enrage which essentially moves half of the rage to being instant. It gives 20 instantly and makes it give only 1 per tick, if I'm reading correctly.

Improved Enrage talent still adds additional up front rage.

sim/druid/enrage.go
```diff
@@ -59,7 +51,7 @@ func (druid *Druid) registerEnrageSpell() {
 				Period:   time.Second * 1,
 				OnAction: func(sim *core.Simulation) {
 					if druid.EnrageAura.IsActive() {
-						druid.AddRage(sim, 1, rageMetrics)
+						druid.AddRage(sim, 2, rageMetrics)
```
balor89, line 54: So this 2 would become a 1, from the SoD tuning aura.

sim/druid/forms.go
```diff
@@ -263,144 +263,126 @@ func (druid *Druid) registerCatFormSpell() {
...
-// 			}
-// 			druid.BearFormAura.Activate(sim)
-// 		},
-// 	})
-// }
+func (druid *Druid) registerBearFormSpell() {
+	actionID := core.ActionID{SpellID: core.TernaryInt32(druid.Level < 40, 5487, 9634)}
+
+	statBonus := druid.GetFormShiftStats().Add(stats.Stats{
+		stats.AttackPower: 3 * float64(druid.Level),
```
balor89, line 270: Trivial, but I recently discovered that Form Stats doesn't use druid.level in game, it actually uses Floor(FeralCombatSkill / 5). So when using the feral combat ring rune, you end up getting +3 AP and + 32 health, or something like that. Same for cat.

We can probably skip this though if it's not a trivial change.

sim/druid/forms.go
```diff
@@ -263,144 +263,126 @@ func (druid *Druid) registerCatFormSpell() {
...
+	feralApDep := druid.NewDynamicStatDependency(stats.FeralAttackPower, stats.AttackPower, 1)
+
+	var hotwDep *stats.StatDependency
+	if druid.Talents.HeartOfTheWild > 0 {
+		hotwDep = druid.NewDynamicMultiplyStat(stats.Stamina, 1.0+0.04*float64(druid.Talents.HeartOfTheWild))
+	}
+
+	sotfdtm := 1.0
+	if druid.HasRune(proto.DruidRune_RuneChestSurvivalOfTheFittest) {
+		sotfdtm = 0.80
```
balor89, line 283: So technically, [Survival of the Fittest](https://www.wowhead.com/classic/spell=411115/survival-of-the-fittest) works as two separate -10% damage modifiers. The first is always active, the second is only active in form.

So for bear, it's actually 0.9*0.9 = 0.81 damage mod from the rune.

sim/druid/item_sets_pve.go
```diff
@@ -295,11 +295,13 @@ var ItemSetCenarionRage = core.NewItemSet(core.ItemSet{
...
 		4: func(agent core.Agent) {
-			// TODO: Enrage
+			druid := agent.(DruidAgent).GetDruid()
+			druid.Enrage.CD.Duration -= time.Second * 30
 		},
 		// Bear Form and Dire Bear Form increase all threat you generate by an additional 20%, and Cower now removes all your threat against the target but has a 20 sec longer cooldown.
 		6: func(agent core.Agent) {
-			// TODO: Bear, Dire Bear forms
+			druid := agent.(DruidAgent).GetDruid()
+			druid.PseudoStats.ThreatMultiplier += .2
```
kayla-glick, line 304: This is going to apply all the time rather than just in bear form. Maybe we can use an attribute on the druid to track the threat value?

sim/druid/item_sets_pve.go
```diff
@@ -436,6 +438,9 @@ var ItemSetCunningOfStormrage = core.NewItemSet(core.ItemSet{
 			druid.RegisterAura(core.Aura{
 				Label: "S03 - Item - T2- Druid - Feral 4P Bonus",
 				OnInit: func(aura *core.Aura, sim *core.Simulation) {
+					if druid.form != Cat {
+						return
+					}
```
kayla-glick, line 443: Since forms can be changed mid-combat I don't think we want to add this restriction

sim/druid/item_sets_pve.go
```diff
@@ -485,12 +490,80 @@ var ItemSetFuryOfStormrage = core.NewItemSet(core.ItemSet{
...
+			// results := make([]*core.SpellResult, numHits)
+
+			// core.MakePermanent(druid.RegisterAura(core.Aura{
+			// 	Label: "S03 - Item - T2 - Druid - Guardian 6P Bonus",
+			// 	OnSpellHitDealt: func(aura *core.Aura, sim *core.Simulation, spell *core.Spell, result *core.SpellResult) {
+			// 		if spell.SpellCode == SpellCode_DruidSwipeBear {
+			// 			// TODO: Troubleshoot - Saeyon
+			// 			for _, cleaveMob := range results {
+			// 				druid.LacerateBleed.Cast(sim, cleaveMob.Target)
+			// 				druid.LacerateBleed.Dot(cleaveMob.Target).SetStacks(sim, druid.LacerateBleed.Dot(result.Target).GetStacks())
```
kayla-glick, line 562: A bear friend told me that the lacerate spreads at the original dot duration as well so we'd want to set both stacks and duration here

sim/druid/item_sets_pve.go
```diff
@@ -485,12 +490,80 @@ var ItemSetFuryOfStormrage = core.NewItemSet(core.ItemSet{
...
+				ProcMask:    core.ProcMaskEmpty,
+				Flags:       core.SpellFlagMeleeMetrics | core.SpellFlagNoOnCastComplete | core.SpellFlagPassiveSpell,
+
+				DamageMultiplier: 1,
+				ThreatMultiplier: 1,
+
+				ApplyEffects: func(sim *core.Simulation, target *core.Unit, spell *core.Spell) {
+					spell.CalcAndDealDamage(sim, target, curDmg, spell.OutcomeAlwaysHit)
+				},
+			})
```
kayla-glick, line 513: It looks like this spell modifier the `jump target` of Maul which means it just causes Maul to hit a second target, so I think you just want to cast Maul at the second target rather than creating a brand new damaging spell

sim/druid/item_sets_pve.go
```diff
@@ -485,12 +490,80 @@ var ItemSetFuryOfStormrage = core.NewItemSet(core.ItemSet{
...
+			core.MakePermanent(druid.RegisterAura(core.Aura{
+				Label: "S03 - Item - T2 - Druid - Guardian 2P Bonus",
+				OnSpellHitDealt: func(aura *core.Aura, sim *core.Simulation, spell *core.Spell, result *core.SpellResult) {
+					if result.Landed() && spell.SpellCode == SpellCode_DruidMaul {
+						cleaveAura.Activate(sim)
+						curDmg = result.Damage / result.ResistanceMultiplier
+						cleaveHit.Cast(sim, druid.Env.NextTargetUnit(result.Target))
+						cleaveHit.SpellMetrics[result.Target.UnitIndex].Casts--
+					}
+				},
```
kayla-glick, line 536: If I'm reading the effect properly, this `OnSpellHitDealt` should check for a swipe hit and activate the cleave aura. The cleave aura is the one that should check for maul and cast a second maul if the sim has at least 2 targets

sim/druid/runes.go
```diff
@@ -25,8 +25,8 @@ func (druid *Druid) ApplyRunes() {
...
 	druid.registerSunfireSpell()
+	druid.applyMangle()
 
 	// Belt
 	druid.applyBerserk()
```
kayla-glick, line 32: Looks like the Bear Berserk effect is still missing

sim/druid/frenzied_regeneration.go
```diff
@@ -0,0 +1,102 @@
...
+
+					if druid.FrenziedRegenerationAura.IsActive() {
+						druid.SpendRage(sim, rageDumped, rageMetrics)
+						druid.GainHealth(sim, healthGained, healthMetrics)
+					}
+				},
+			})
+
+			druid.FrenziedRegenerationAura.Activate(sim)
+		},
```
kayla-glick, line 100: I think we still need to add handling for the improved frenzied regen since that' what Bears typically run

sim/druid/item_sets_pve.go
```diff
@@ -683,11 +757,50 @@ var ItemSetGenesisFury = core.NewItemSet(core.ItemSet{
...
+				OnStacksChange: func(aura *core.Aura, sim *core.Simulation, oldStacks, newStacks int32) {
+					druid.MangleBear.DamageMultiplierAdditive += 0.1 * float64(newStacks-oldStacks)
+					druid.SwipeBear.DamageMultiplierAdditive += 0.1 * float64(newStacks-oldStacks)
+				},
+			})
+
+			druid.Tank2PieceAqAura = druid.RegisterAura(core.Aura{
+				Label:           "S03 - Item - TAQ - Druid - Guardian 2P Bonus",
+				ActionID:        core.ActionID{SpellID: 1213188},
+				ActionIDForProc: core.ActionID{SpellID: 1213190},
```
kayla-glick, line 775: I'm not 100% sure what `ActionIDForProc` is used for but generally we include the action ID on the proc aura (the one above this one) so that users can track that aura on the timeline / APL / buffs tab then leave it empty on a trigger aura like this unless there's a need to make APL decisions based on having the bonus

sim/druid/mangle.go
```diff
@@ -35,22 +45,33 @@ func (druid *Druid) registerMangleBearSpell() {
...
+				dotBonusCrit = druid.FuryOfStormrageCritRatingBonus
+			}
 
+			spell.BonusCritRating += dotBonusCrit
 			result := spell.CalcAndDealDamage(sim, target, baseDamage, spell.OutcomeMeleeSpecialHitAndCrit)
+			spell.BonusCritRating -= dotBonusCrit
 
 			if result.Landed() {
 				mangleAuras.Get(target).Activate(sim)
+				if stacks := int32(druid.GetStat(stats.Defense)); stacks > 0 {
```
balor89, line 69: I'm assuming that stats.Defense here is +Defense you have, and not your literal Defense Stat? Since the in-game defense stat is usually 300 baseline and Defender's Resolve looks at Defense > 300.

Just want to make sure we're not adding 1200 extra AP, lol.

sim/druid/mangle.go
balor89: How does DamageMultiplierAdditive work under the hood? Since as far as I know, Savage Fury is a flat 1.2x mod on all mangle damage done.

sim/druid/swipe.go
```diff
@@ -5,31 +5,18 @@ import (
...
-		25: 2,
-		40: 3,
-		50: 4,
-		60: 6,
-	}[druid.Level]
-
-	level := SwipeLevel[rank]
-	spellID := SwipeSpellId[rank]
-	baseDamage := SwipeBaseDamage[rank]
+	baseDamage := 83 + .1*druid.GetStat(stats.AttackPower)
```
balor89, line 19: Why were the level-based ranks removed? Seems like there's no reason to have removed them if they were already in.

balor89 (commented):
This isn't a review of the code correctness or quality, just of the accuracy around in-game numbers and mechanics.

I had two stand alone comments before I started making them as part of a review, so check those out too.

sim/druid/mangle.go
emsimpson92: Ra informed me that stats.Defense is your def - 300

sim/druid/swipe.go
emsimpson92: good catch

sim/druid/item_sets_pve.go
emsimpson92: This doesn't even work at the moment but I'll leave a comment for this

sim/druid/runes.go
emsimpson92: The cleave isn't implemented yet, but the rage reduction on lacerate and CD refresh have been. They are in lacerate.go and mangle.go

sim/druid/lacerate.go
```diff
@@ -0,0 +1,131 @@
...
+			},
+			IgnoreHaste: true,
+		},
+
+		DamageMultiplier: 1,
+		ThreatMultiplier: 3.33,
+		// TODO: Berserk 3 target lacerate cleave - Saeyon
+
+		ApplyEffects: func(sim *core.Simulation, target *core.Unit, spell *core.Spell) {
+			baseDamage := spell.Unit.MHWeaponDamage(sim, spell.MeleeAttackPower()) * (.2 * float64(druid.LacerateBleed.Dot(target).GetStacks()) * initialDamageMul)
```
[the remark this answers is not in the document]
emsimpson92: the stack is applied before dealing direct damage

sim/druid/swipe.go
emsimpson92: I saw swipe was a rune and thought that it was new in SOD. Turns out it's just swipe cat that is new. I've added level scaling back

sim/druid/lacerate.go
```diff
@@ -0,0 +1,131 @@
...
+			}
+		},
+	})
+}
+
+func (druid *Druid) registerLacerateBleedSpell() {
+	if !druid.HasRune(proto.DruidRune_RuneLegsLacerate) {
+		return
+	}
+	tickDamage := 20.0
```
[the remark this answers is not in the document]
emsimpson92: I think that may have come from some wotlk code. Not sure

sim/druid/item_sets_pve.go
emsimpson92: The main difference is that afaik cleave effects like this can't be independently parried, so this cleave hit has spell.OutcomeAlwaysHit

sim/common/vanilla/item_effects.go
```diff
@@ -2371,13 +2375,28 @@ func init() {
...
+	})
+	core.NewItemEffect(RazorbrambleShoulderpads, func(agent core.Agent) {
+		DamageShieldWithThreatMod(agent.GetCharacter(), 1213816, 80, 2, "Damage Shield Razorbramble Shoulderpads")
+	})
+	core.NewItemEffect(RazorbrambleCowl, func(agent core.Agent) {
+		DamageShieldWithThreatMod(agent.GetCharacter(), 1213813, 100, 2, "Damage Shield Razorbramble Cowl")
+	})
+	core.NewItemEffect(LodestoneofRetaliation, func(agent core.Agent) {
+		DamageShieldWithThreatMod(agent.GetCharacter(), 1213816, 80, 2, "Damage Shield Lodestone of Retaliation")
+	})
```
kayla-glick, line 2398: I didn't realize these were included here but I already implemented these in https://github.com/wowsims/sod/blob/master/sim/common/sod/item_effects/phase_6.go the other day while adding in all of the missing Thorns effects

ui/core/launched_sims.ts
```diff
@@ -33,8 +33,8 @@ export const simLaunchStatuses: Record<Spec, SimStatus> = {
 		status: LaunchStatus.Alpha,
 	},
 	[Spec.SpecFeralTankDruid]: {
-		phase: Phase.Phase1,
-		status: LaunchStatus.Unlaunched,
+		phase: Phase.Phase5,
+		status: LaunchStatus.Alpha,
```
kayla-glick, line 37: Should we go ahead and update this to Phase 6? Also in `index.html`?

sim/druid/tank/tank.go
```diff
@@ -72,12 +76,16 @@ func (bear *FeralTankDruid) GetDruid() *druid.Druid {
 
 func (bear *FeralTankDruid) Initialize() {
 	bear.Druid.Initialize()
-	bear.RegisterFeralTankSpells()
+	queuedRealismICD := &core.Cooldown{
+		Timer:    bear.NewTimer(),
+		Duration: core.SpellBatchWindow * 10,
+	}
```
kayla-glick, line 82: ❓ Since this is only use for Maul should we just initialize this inside of the Maul code?

sim/druid/item_sets_pve.go
```diff
@@ -1,132 +1 @@
...
-		3: func(agent core.Agent) {
-			c := agent.GetCharacter()
-			c.AddStat(stats.Stamina, 10)
-		},
-		6: func(agent core.Agent) {
-			c := agent.GetCharacter()
-			c.AddStat(stats.HealingPower, 22)
-		},
-	},
-})
```
kayla-glick, line 132: Were these removed by mistake?

sim/druid/item_sets_pve_phase_4.go
```diff
@@ -225,7 +226,8 @@ var ItemSetCenarionRage = core.NewItemSet(core.ItemSet{
 		},
 		// Reduces the cooldown of Enrage by 30 sec and it no longer reduces your armor.
 		4: func(agent core.Agent) {
-			// TODO: Enrage
+			druid := agent.(DruidAgent).GetDruid()
+			druid.Enrage.CD.Duration -= time.Second * 30
```
kayla-glick, line 230: This changed since you put up the PR but we should modify the new CD `FlatModifier` field for effects like this now. CDs store the original Duration, a cumulative flat modifier like this one, and then a cumulative multiplier, then the sim calculates the real duration from those like the game does

```suggestion
			druid.Enrage.CD.FlatModifier -= time.Second * 30
```

sim/druid/item_sets_pve_phase_6.go
```diff
@@ -201,17 +219,10 @@ func (druid *Druid) applyTAQGuardian4PBonus() {
...
-	}
-
-	druid.RegisterAura(core.Aura{
-		Label: label,
-		OnInit: func(aura *core.Aura, sim *core.Simulation) {
-			// TODO
-		},
+	druid.OnSpellRegistered(func(spell *core.Spell) {
+		if spell.SpellCode == SpellCode_DruidMangleBear {
+			spell.CD.Duration -= 1500 * time.Millisecond
```
kayla-glick, line 224:
```suggestion
			spell.CD.FlatModifier-= 1500 * time.Millisecond
```

sim/druid/item_sets_pve_phase_6.go
```diff
@@ -183,15 +183,33 @@ var ItemSetGenesisFury = core.NewItemSet(core.ItemSet{
 
 // Each time you Dodge while in Dire Bear Form, you gain 10% increased damage on your next Mangle or Swipe, stacking up to 5 times.
 func (druid *Druid) applyTAQGuardian2PBonus() {
-	label := "S03 - Item - TAQ - Druid - Guardian 2P Bonus"
-	if druid.HasAura(label) {
-		return
-	}
```
kayla-glick, line 189: This will end up breaking if using the set bonus + corresponding should rune with this change. We should leave the `HasAura` check to prevent that

sim/druid/druid.go
```diff
@@ -121,13 +130,17 @@ type Druid struct {
 	SolarEclipseProcAura     *core.Aura
 	LunarEclipseProcAura     *core.Aura
 	WildStrikesBuffAura      *core.Aura
+	Tank2PieceAqAura         *core.Aura
+	Tank2PieceAqProcAura     *core.Aura
```
kayla-glick, line 134: Are these two auras used outside of the set bonus code? If not you should be able to remove these and just store them as locals

sim/druid/item_sets_pve_phase_4.go
```diff
@@ -236,17 +238,7 @@ var ItemSetCenarionRage = core.NewItemSet(core.ItemSet{
...
-	}
-
-	druid.RegisterAura(core.Aura{
-		Label: label,
-		OnInit: func(aura *core.Aura, sim *core.Simulation) {
-			// TODO
-		},
-	})
+	druid.CenarionRageThreatBonus = .2
 }
```
kayla-glick, line 242: This will double dip without the aura code here. What you can do for this bonus is in the `OnInit`, grab the old `OnGain` and `OnExpire` for bear form, then override them, calling the old ones and then also adding the threat mod here:
```
OnInut: func(aura *core.Aura, sim *core.Simulation) {
  oldOnGain := druid.BearFormAura.OnGain
  druid.BearFormAura.OnGain = func(aura *core.Aura, sim *core.Simulation) {
    oldOnGain(aura, sim)
    druid.Pseudostats.ThreatMultiplier += 0.20
  }

 oldOnExpire := druid.BearFormAura.OnExpire
 druid.BearFormAura.OnExpire = func(aura *core.Aura, sim *core.Simulation) {
   oldOnExpire(aura, sim)
   druid.Pseudostats.ThreatMultiplier -= 0.20
 }
}
```

sim/druid/item_sets_pve_phase_4.go
kayla-glick: Oh also since this is also a rune, we'll want to follow the pattern of creating a function that does the aura check / creates an aura to handle the set bonus. Following the other example, create the aura, then in the `OnInit` you can make the change to Enrage's CD FlatModifier

sim/druid/item_sets_pve_phase_5.go
```diff
@@ -237,55 +236,76 @@ var ItemSetFuryOfStormrage = core.NewItemSet(core.ItemSet{
 
 // Swipe(Bear) also causes your Maul to hit 1 additional target for the next 6 sec.
 func (druid *Druid) applyT2Guardian2PBonus() {
-	label := "S03 - Item - T2 - Druid - Guardian 2P Bonus"
-	if druid.HasAura(label) {
+	if druid.Env.GetNumTargets() == 1 {
 		return
 	}
```
kayla-glick, line 241: We'll also want to keep this aura check but it can go after the target count check, that way we don't double dip (I think it would just error from the same spell registered twice)

sim/druid/item_sets_pve_phase_5.go
```diff
@@ -237,55 +236,76 @@ var ItemSetFuryOfStormrage = core.NewItemSet(core.ItemSet{
...
-	if druid.HasAura(label) {
-		return
-	}
-
-	druid.RegisterAura(core.Aura{
-		Label: label,
-		OnInit: func(aura *core.Aura, sim *core.Simulation) {
-			// TODO
-		},
-	})
```
kayla-glick, line 269: This also needs to have the aura stuff here because of the tier bonus + shoulder runes, so you could move this bonus crit rating into the `OnInit` here and that should be fine.

Another way we could do this is to override the `ApplyEffects` of the affected spells so that before calling the old one we check if lacerate is active then add 5% bonus crit to the spell, call the original `ApplyEffects`, then remove the bonus crit. Basically what you did in the spell code but all in one place.

You would also want to add a check in this function that returns if the druid doesn't have the Lacerate rune

```
for _, spell := range []*DruidSpell{druid.MangleBear, druid.SwipeBear, druid.Maul, druid.Lacerate} {
  if spell == nil {
    continue
  }

  oldApplyEffects := spell.ApplyEffects
  spell.ApplyEffects = func(sim *core.Simulation, target *core.Unit, spell *core.Spell) {
    bonusCrit := 0.0
    if druid.LacerateBleed.Dot(target).GetStacks() > 0 {
      bonusCrit = 5 * core.CritRatingPerCritChance
    }

    spell.BonusCritRating += bonusCrit
    oldApplyEffects(sim, target, spell)
    spell.BonusCritRating -= bonusCrit
  }
}
```

sim/druid/item_sets_pve_phase_5.go
```diff
@@ -237,55 +236,76 @@ var ItemSetFuryOfStormrage = core.NewItemSet(core.ItemSet{
...
```
kayla-glick, line 288: This stuff also needs to come back but bumping this set bonus as incomplete

sim/druid/item_sets_pve_phase_6.go
```diff
@@ -183,15 +183,33 @@ var ItemSetGenesisFury = core.NewItemSet(core.ItemSet{
...
+	druid.Tank2PieceAqProcAura = druid.RegisterAura(core.Aura{
+		Label:     "Guardian 2P Bonus Proc",
+		ActionID:  core.ActionID{SpellID: 1213188},
+		Duration:  time.Second * 10,
+		MaxStacks: 5,
+		OnStacksChange: func(aura *core.Aura, sim *core.Simulation, oldStacks, newStacks int32) {
+			druid.MangleBear.DamageMultiplierAdditive += 0.1 * float64(newStacks-oldStacks)
```
kayla-glick, line 192: This would need a guard checking `if druid.MangleBear != nil` to prevent erroring if not using the Mangle rune

sim/druid/item_sets_pve_phase_6.go
```diff
@@ -183,15 +183,33 @@ var ItemSetGenesisFury = core.NewItemSet(core.ItemSet{
...
+				druid.Tank2PieceAqProcAura.Activate(sim)
+				druid.Tank2PieceAqProcAura.AddStack(sim)
+			}
+		},
+		OnSpellHitDealt: func(aura *core.Aura, sim *core.Simulation, spell *core.Spell, result *core.SpellResult) {
+			if spell.SpellCode == SpellCode_DruidMangleBear || spell.SpellCode == SpellCode_DruidSwipeBear {
+				druid.Tank2PieceAqProcAura.SetStacks(sim, 0)
+			}
 		},
 	})
```
kayla-glick, line 214: Just simplifying with the `MakePermanent` helper

```suggestion
	core.MakePermanent(druid.Tank2PieceAqAura = druid.RegisterAura(core.Aura{
		Label:    "S03 - Item - TAQ - Druid - Guardian 2P Bonus",
		OnSpellHitTaken: func(aura *core.Aura, sim *core.Simulation, spell *core.Spell, result *core.SpellResult) {
			if druid.form == Bear && spell.ProcMask.Matches(core.ProcMaskMelee) && result.Outcome.Matches(core.OutcomeDodge) {
				druid.Tank2PieceAqProcAura.Activate(sim)
				druid.Tank2PieceAqProcAura.AddStack(sim)
			}
		},
		OnSpellHitDealt: func(aura *core.Aura, sim *core.Simulation, spell *core.Spell, result *core.SpellResult) {
			if spell.SpellCode == SpellCode_DruidMangleBear || spell.SpellCode == SpellCode_DruidSwipeBear {
				druid.Tank2PieceAqProcAura.SetStacks(sim, 0)
			}
		},
	}))
```

sim/druid/item_sets_pve_phase_6.go
kayla-glick: Oh and also this should go inside of the aura's `OnInit` function to work properly with the set bonus / shoulder rune

sim/druid/items.go
```diff
@@ -15,6 +15,7 @@ const (
 	IdolOfFerocity                   = 22397
 	IdolOfTheMoon                    = 23197
 	IdolOfBrutality                  = 23198
+	IdolOfCruelty                    = 232424
```
kayla-glick, line 18: Does this need to be implemented still?

sim/druid/lacerate.go
```diff
@@ -0,0 +1,131 @@
...
+		Cast: core.CastConfig{
+			DefaultCast: core.Cast{
+				GCD: core.GCDDefault,
+			},
+			IgnoreHaste: true,
+		},
+
+		DamageMultiplier: 1,
+		ThreatMultiplier: 3.33,
+		// TODO: Berserk 3 target lacerate cleave - Saeyon
```
kayla-glick, line 44: My suggestion for this would be to bake a target count into lacerate's code and default it to. Maybe you can use an attribute on the druid to track the current target count that lacerate should hit. Sounds like there's two different effects that make lacerate affect multiple targets?

sim/druid/item_sets_pve_phase_5.go
kayla-glick: If you implemented this here then you would want to remove the similar code from the mangle, swipe, lacerate, and maul spells

sim/druid/mangle.go
```diff
@@ -35,22 +46,34 @@ func (druid *Druid) registerMangleBearSpell() {
 			IgnoreHaste: true,
 			CD: core.Cooldown{
 				Timer:    druid.NewTimer(),
-				Duration: time.Duration(float64(time.Second) * (6 - durReduction)),
+				Duration: time.Second * 6,
 			},
 		},
+		// TODO: Berserk 3 target mangle cleave - Saeyon
```
kayla-glick, line 52: Same comment as above with lacerate ^

sim/druid/items.go
emsimpson92: no

PR #1197, opened 2024-12-23
Phase 7 Shoulder "Runes"

kayla-glick:
image


sim/shaman/item_sets_phase_4.go
```diff
@@ -0,0 +1,307 @@
+package shaman
```
kayla-glick, line 1: With these changes we're going to be breaking out a lot of functions so I think it would be good to separate PvE sets by phase at least starting with Phase 4

sim/shaman/item_sets_phase_4.go
```diff
@@ -0,0 +1,307 @@
...
+
+	core.MakePermanent(shaman.RegisterAura(core.Aura{
+		Label: label,
+		OnSpellHitDealt: func(aura *core.Aura, sim *core.Simulation, spell *core.Spell, result *core.SpellResult) {
+			if spell.SpellCode == SpellCode_ShamanLightningBolt && spell.ProcMask.Matches(core.ProcMaskSpellDamage) && result.DidCrit() && sim.Proc(.35, "Power Surge") {
+				shaman.PowerSurgeDamageAura.Activate(sim)
+			}
+		},
+	}))
+}
```
kayla-glick, line 95: Bonus defined as follows:
1. Comment with the spell tooltip description
2. Class method `apply[Tier#][SpecName][#Pieces]Bonus()`
3. Define the label then return if the character already has the aura registered. This is more important for more complex bonuses where `GetOrRegisterAura` may not necessarily prevent other side effects and is a more reliable pattern we can follow

sim/shaman/runes.go
```diff
@@ -48,6 +51,29 @@ func (shaman *Shaman) ApplyRunes() {
...
+	case int32(proto.ShamanRune_RuneShouldersChieftain):
+		shaman.applyT1Enhancement6PBonus()
+	case int32(proto.ShamanRune_RuneShouldersWindwalker):
+		shaman.applyT1Tank2PBonus()
+	case int32(proto.ShamanRune_RuneShouldersShieldMaster):
+		shaman.applyT1Tank4PBonus()
+	case int32(proto.ShamanRune_RuneShouldersTotemicProtector):
+		shaman.applyT1Tank6PBonus()
+	}
+}
```
kayla-glick, line 75: Should runes get a method inside of  `runes.go` with a switch statement. I ordered these by Tier #, then Spec, then # of pieces

sim/core/aura.go
```diff
@@ -18,6 +18,7 @@ type OnInit func(aura *Aura, sim *Simulation)
 type OnReset func(aura *Aura, sim *Simulation)
 type OnDoneIteration func(aura *Aura, sim *Simulation)
 type OnGain func(aura *Aura, sim *Simulation)
+type OnRefresh func(aura *Aura, sim *Simulation)
```
kayla-glick, line 21: New callback used to handle some effects that we can't quite handle through a buff aura, e.g. Shield Mastery restoring mana on apply/refresh

PR #1202, opened 2024-12-27
add pally souls implementations

Adamrch:


sim/paladin/item_sets_pve.go
```diff
@@ -8,6 +8,38 @@ import (
...
+	PaladinT2Ret2P   = 467518
+	PaladinT2Ret4P   = 467526
+	PaladinT2Ret6P   = 467529
+	PaladinTAQRet2P  = 1213397
+	PaladinTAQRet4P  = 1213406
+	PaladinZG2P      = 468401
+	PaladinZG3P      = 468428
+	PaladinZG5P      = 468431
+	PaladinRAQ3P     = 123467
+)
```
kayla-glick, line 41: Can these be deleted now?

sim/paladin/item_sets_pve.go
```diff
@@ -443,78 +204,482 @@ var ItemSetWilfullJudgement = core.NewItemSet(core.ItemSet{
...
-		3: func(agent core.Agent) {
-			paladin := agent.(PaladinAgent).GetPaladin()
+		3: func(agent core.Agent) { agent.(PaladinAgent).GetPaladin().applyPaladinRAQ3P() },
+	},
+})
+
+func (paladin *Paladin) applyPaladinT1Prot2P() {
+	// (2) Set: Increases the block value of your shield by 30.
+	paladin.AddStat(stats.BlockValue, 30)
+}
```
kayla-glick, line 223: Won't this bonus stack if you wear the enchant + set bonus? You could apply it with an aura like this:
https://github.com/wowsims/sod/blob/4b20b13717962d5075f47df57f64df8ddf3cbb4d/sim/shaman/item_sets_pve_phase_4.go#L163-L192

sim/paladin/item_sets_pve.go
```diff
@@ -443,78 +204,482 @@ var ItemSetWilfullJudgement = core.NewItemSet(core.ItemSet{
...
+	paladin := agent.(PaladinAgent).GetPaladin()
+
+	if paladin.GetAuraByID(core.ActionID{SpellID: bonusID}) != nil {
+		return true // Do not apply bonus aura more than once (Due to Soul)
+	}
+
+	paladin.RegisterAura(core.Aura{
+		Label:    bonusString,
+		ActionID: core.ActionID{SpellID: bonusID},
+	})
```
kayla-glick, line 682: I'm not a big fan of how this creates a separate aura from other set bonus auras created in the individual bonus functions. It ends up creating duplicates vs the way that I just used labels in other classes' code
https://github.com/wowsims/sod/blob/4b20b13717962d5075f47df57f64df8ddf3cbb4d/sim/shaman/item_sets_pve_phase_4.go#L164-L175

sim/paladin/item_sets_pve.go
Adamrch: I'm using them for the auras to track duplicates.

sim/paladin/item_sets_pve.go
Adamrch: it would, but I went off of the assumption that stat stick auras would stack and unique effect auras would not. We still don't know exactly how it will work but I think it is a reasonable assumption.

PR #1218, opened 2025-01-17
Ret preset updates and off-GCD fix for strict sequences

hillerstorm:
I also found some cases where multiple MakePermanent were run on the same aura, which just basically ment Activate were run multiple times on reset.
This was most noticeable on Sanctity Aura, which got re-permad on reset, so adding that aura to the test cases gave extreme diffs in multi-core-comparisons :)

This PR will make twisting sim close to @Adamrch branch. After doing some comparisons, that branch gets a tiny bit more extra attacks on average which is weird but could be caused by all kinds of stuff :D


sim/core/buffs.go
```diff
@@ -2211,16 +2220,6 @@ func CreateExtraAttackAuraCommon(character *Character, buffActionID ActionID, au
...
 
-	MakePermanent(character.GetOrRegisterAura(Aura{
-		Label:     "Extra Attacks  (Main Hand)", // Tracks Stored Extra Attacks from all sources
-		ActionID:  ActionID{SpellID: 21919},     // Thrash ID
-		Duration:  NeverExpires,
-		MaxStacks: 4, // Max is 4 extra attacks stored - more can proc after
-		OnInit: func(aura *Aura, sim *Simulation) {
-			aura.Unit.AutoAttacks.mh.extraAttacksAura = aura
-		},
-	}))
```
kayla-glick, line 2222: Deleting this may cause problems and might be why you're seeing differences in extra attacks. Should we just remove `MakePermanent` here and set the Duration + OnReset instead?
```
character.GetOrRegisterAura(Aura{
		Label:     "Extra Attacks  (Main Hand)", // Tracks Stored Extra Attacks from all sources
		ActionID:  ActionID{SpellID: 21919},     // Thrash ID
		Duration:  NeverExpires,
		MaxStacks: 4, // Max is 4 extra attacks stored - more can proc after
    OnReset: func(aura *Aura, sim *Simulation) {
      aura.activate(sim)
    },
		OnInit: func(aura *Aura, sim *Simulation) {
			aura.Unit.AutoAttacks.mh.extraAttacksAura = aura
		},
	}))
```

sim/paladin/paladin.go
```diff
@@ -232,7 +234,11 @@ func (paladin *Paladin) registerStopAttackMacros() {
 func (paladin *Paladin) ResetCurrentPaladinAura() {
 	paladin.currentPaladinAura = nil
 	if paladin.primaryPaladinAura == proto.PaladinAura_SanctityAura {
-		paladin.currentPaladinAura = core.SanctityAuraAura(paladin.GetCharacter())
+		if paladin.sanctityAura == nil {
+			paladin.sanctityAura = core.SanctityAuraAura(paladin.GetCharacter())
+		}
```
kayla-glick, line 239: We should probably just set `paladin.sanctityAura` in the initializer instead of here so that we don't have to check if it's `nil` like this

sim/paladin/protection/protection.go
```diff
@@ -30,7 +30,6 @@ func NewProtectionPaladin(character *core.Character, options *proto.Player) *Pro
 
 	prot := &ProtectionPaladin{
 		Paladin:                         pal,
-		primarySeal:                     protOptions.PrimarySeal,
```
kayla-glick, line 33: Is removing this going to cause the prot paladin to use the wrong primary seal?

ui/retribution_paladin/presets.ts
```diff
@@ -67,37 +67,38 @@ export const Phase2RetSoMGear = PresetUtils.makePresetGear('P2 SoM', Phase2RetSo
...
 	customCondition: player => player.getLevel() == 60,
 });
 
 export const GearPresets = {
 	[Phase.Phase1]: [Phase1RetGear],
 	[Phase.Phase2]: [Phase2RetSoCGear, Phase2RetSoMGear],
 	[Phase.Phase3]: [Phase3RetSoMGear],
-	[Phase.Phase4]: [Phase4RetTwistGear, Phase4RetTwist6pT1Gear, Phase4RetExodinGear, Phase4RetExodin6pT1Gear],
-	[Phase.Phase5]: [Phase5TwistingGear, Phase5ExodinGear, Phase5ShockadinGear],
+	[Phase.Phase4]: [],
```
kayla-glick, line 96: Looks like the Phase 4 sets are missing now?

ui/retribution_paladin/presets.ts
```diff
@@ -112,44 +113,39 @@ export const APLP2Ret = PresetUtils.makePresetAPLRotation('P2 Ret/Shockadin', AP
...
 	customCondition: player => player.getLevel() === 60,
 });
 
 export const APLPresets = {
 	[Phase.Phase1]: [APLP1Ret],
 	[Phase.Phase2]: [APLP2Ret],
 	[Phase.Phase3]: [APLP3Ret],
-	[Phase.Phase4]: [APLP4RetTwist, APLP4RetTwist6pT1, APLP4RetExodin, APLP4RetExodin6pT1],
-	[Phase.Phase5]: [APLPP5Twisting4DRSlow, APLPP5Twisting4DRSlower, APLPP5Exodin, APLPP5Shockadin],
+	[Phase.Phase4]: [],
```
kayla-glick, line 139: Looks like the Phase 4 APLs are missing now?

sim/core/buffs.go
hillerstorm: Nah, I saw the diff long before I removed this (and the diff was only 0.5 extra attacks on 100k iterations).
I added logging in the OnInit here and it never got ran so all this whole block does is add another `MakePermanent` on the Thrash-auras here https://github.com/wowsims/sod/blob/b3b7861ef61ecd8108eb92feca15dc610da9df6a/sim/core/attack.go#L565

sim/paladin/protection/protection.go
hillerstorm: There is already a `primarySeal` on the Paladin type itself that gets set during Initialize, so the prot/ret specific ones were never used. It's the same options field used in both all cases and the spec-specific ones were just redundant :)

ui/retribution_paladin/presets.ts
hillerstorm: Yea the list was getting too big in the UI... feels like enough to just show current+previous phase just like Rogues did with APLs?
Sure, I can keep them here in the preset dict but showing 11 different preset gear options and 10 different APLs is... a bit much imo :D
especially when 4 of those are from *two* phases back

ui/retribution_paladin/presets.ts
hillerstorm: See thoughts from comment on gear sets.

Also, the twisting APLs would've been removed anyways and the P4 Exodin6pT1 is just seal stacking so they could've just been condensed into Phase5 APLs.

ui/retribution_paladin/presets.ts
hillerstorm: I can add the relevant ones back just for completeness sake, but when P7 sets come, we will have to clean this up :)

ui/retribution_paladin/presets.ts
hillerstorm: Adding the relevant APLs back and trimming away those that can use later versions

sim/paladin/protection/protection.go
kayla-glick: Oh gotcha!

ui/retribution_paladin/presets.ts
kayla-glick: No that's fine totally get that

PR #1230, opened 2025-01-19
Item Swap - Weapon slots & Aura helpers

1337LutZ:
Initial Item Swap migration from the Cata repo. Working on the Weapon slots first.

**Todo**
- [x] Migrate (full gear) Item Swap logic and add limited slot panic for now
- [x] Migrate PPM to DPM (Dynamic Proc Manager)
- [x] Refactor UI to use the new `items` input for future full gear item swap support
- [x] Rewrite `aura.Unit.Env.MeasuringStats && aura.Unit.Env.State != Finalized > aura.Unit.AddStat` logic to `EnableBuildPhaseStatDep`
- [x] Migrate all Weapon/Enchant effects


1337LutZ:
The current result changes are due to no longer initializing the same item effects 2x, effectively creating a new RNG seed.

1337LutZ:
@hillerstorm @kayla-glick If you could start reviewing these changes that would be great :D
The small result updates are due to RNG seed changes because the Proc is now done earlier in the code in the aura_helpers instead of a manual handler.

I think there's some cases to validate:
- MH+OH slow > MH+OH quick - See how this affects PPM etc
- MH+OH > 2H and reverse - See how this affects PPM etc

proto/common.proto
```diff
@@ -1151,9 +1151,11 @@ message CustomSpell {
 }
 
 message ItemSwap {
-	ItemSpec mh_item = 1;
-	ItemSpec oh_item = 2;
-	ItemSpec ranged_item = 3;
+	ItemSpec mh_item = 1 [deprecated=true];
+	ItemSpec oh_item = 2 [deprecated=true];
+	ItemSpec ranged_item = 3 [deprecated=true];
```
kayla-glick, line 1156: We can use the `reserved` keyword to truly deprecate these with backwards-compatibility:
```
reserved 1, 2, 3;
resserved "mh_item", "oh_item", "ranged_item";
```

sim/common/itemhelpers/weaponprocs.go
```diff
@@ -55,89 +55,82 @@ func CreateWeaponProcDamage(itemId int32, itemName string, ppm float64, spellId
...
-				if result.Landed() && ppmm.Proc(sim, spell.ProcMask, aura.Label) {
-					procSpell.Cast(sim, result.Target)
-				}
+
+		aura := core.MakeProcTriggerAura(&character.Unit, core.ProcTrigger{
+			Name:              fmt.Sprintf("%s Proc Aura", itemName),
+			Callback:          core.CallbackOnSpellHitDealt,
+			Outcome:           core.OutcomeLanded,
+			DPM:               character.AutoAttacks.NewDynamicProcManagerForWeaponEffect(itemID, ppm, 0),
+			DPMProcType:       core.DPMProcNoWeaponSpecials,
```
kayla-glick, line 64: Wait I thought weapons did proc from specials?

sim/core/apl_actions_misc.go
```diff
@@ -238,15 +238,17 @@ func (action *APLActionItemSwap) IsReady(sim *Simulation) bool {
...
-	if sim.Log != nil {
-		action.character.Log(sim, "Item Swap to set %s", action.swapSet)
-	}
-
-	if action.swapSet == proto.APLActionItemSwap_Main {
-		action.character.ItemSwap.reset(sim)
+	if action.character.ItemSwap.swapSet == action.swapSet {
+		if sim.Log != nil {
+			action.character.Log(sim, "Item Swap already set to %s", action.swapSet)
+		}
```
kayla-glick, line 244: Should we just break after this?

sim/common/itemhelpers/weaponprocs.go
1337LutZ: I've renamed this to make it a bit more clear.
Currently the default for `ProcTriggerAura` is `dpm.ProcWithWeaponSpecials()`. 
However not all procs that could be simplified to a `ProcTriggerAura` use the `dpm.Proc()`, this property should be used to indicate this nuance.

sim/core/apl_actions_misc.go
1337LutZ: We could, however the `SwapItems` call will do the check as well and that way we have that logic all in one place

PR #1252, opened 2025-01-26
Feral tank sim implementation

ironclad88:


sim/druid/item_sets_pve_phase_4.go
```diff
@@ -240,9 +237,54 @@ var ItemSetCenarionRage = core.NewItemSet(core.ItemSet{
...
+		Label:    label,
+		OnInit: func(aura *core.Aura, sim *core.Simulation) {
+			druid.Enrage.CD.FlatModifier -= time.Second * 30
+
+			oldOnGain := druid.EnrageAura.OnGain
+			druid.EnrageAura.OnGain = func(aura *core.Aura, sim *core.Simulation) {
+				preArmor := druid.GetStat(stats.Armor)
+				oldOnGain(aura, sim)
+				postArmor := druid.GetStat(stats.Armor)
+				druid.AddStatDynamic(sim, stats.Armor, preArmor-postArmor)
```
kayla-glick, line 258: Rather than adding stats dynamically here, you could make the enrage armor reduction an attribute on the Druid struct, then set it to 0 in this `OnInit`. That should prevent you from having to override the `OnGain` like this

sim/druid/item_sets_pve_phase_4.go
```diff
@@ -240,9 +237,47 @@ var ItemSetCenarionRage = core.NewItemSet(core.ItemSet{
...
+	}
+
+	core.MakePermanent(druid.RegisterAura(core.Aura{
+		ActionID: core.ActionID{SpellID: 456332}, // Tracking in APL
+		Label:    label,
+		OnInit: func(aura *core.Aura, sim *core.Simulation) {
+			oldOnGain := druid.BearFormAura.OnGain
+			druid.BearFormAura.OnGain = func(aura *core.Aura, sim *core.Simulation) {
+				oldOnGain(aura, sim)
+				druid.PseudoStats.ThreatMultiplier += 0.20
```
kayla-glick, line 271: Threat mods are always multiplicative in-game, so adding and subtracting here is a problem. That's actually an error in `forms.go`, too but it shouldn't be hard to fix.

https://www.wowhead.com/classic/spell=21178/bear-form-passive2 is the 30% threat mod (x1.30) that applies when you go into bear form. The way it interacts with this set bonus and the Feral instinct talent is they add to the bonus that gets multiplied by that passive. So for example, with this set and 5/5 Feral Instinct, the `forms.go` threat should do this:
```
druid.Pseudostats.ThreatMultiplier *= 1.3 (base) + 0.03 * float64(druid.Talents.FeralInstinct) + 0.20 (from this bonus
```

What I think you should do is in `stances.go`, assign an attribute to the druid `bearFormThreatMultiplier` and assign it to `1.3 + 0.03 * float64(druid.Talents.FeralInstinct)`

Then in this set bonus code in the `OnInit`, do `druid.bearFormThreatMultiplier += 0.20 to add the 20%

sim/druid/item_sets_pve_phase_5.go
```diff
@@ -295,32 +264,61 @@ func (druid *Druid) applyT2Guardian2PBonus() {
...
-	// 			}
-	// 		}
-	// 	},
-	// }))
+	if !druid.HasRune(proto.DruidRune_RuneLegsLacerate) {
+		return
+	}
+	druid.FuryOfStormrageLacerateSpread = true
+	core.MakePermanent(druid.RegisterAura(core.Aura{
+		Label: "S03 - Item - T2 - Druid - Guardian 6P Bonus",
```
kayla-glick, line 307: This is missing the `HasAura` check to make sure the bonus doesn't get duplicated when using the gear + shoulder run

sim/druid/item_sets_pve_phase_6.go
```diff
@@ -237,11 +240,6 @@ func (druid *Druid) applyTAQGuardian4PBonus() {
 			spell.CD.FlatModifier -= 1500 * time.Millisecond
 		}
 	})
-	druid.OnInit(func(spell *core.Spell) {
-		if spell.SpellCode == SpellCode_DruidMangleBear {
-			spell.CD.FlatModifier -= 1500 * time.Millisecond
-		}
-	})
```
kayla-glick, line 244: For this bonus we need to create an aura and use the `HasAura` approach in order to not apply the bonus twice. You should be able to get rid of the `OnSpellRegistered`

sim/druid/lacerate.go
```diff
@@ -41,28 +41,33 @@ func (druid *Druid) registerLacerateSpell() {
...
-				if hasGore && sim.Proc(0.15, "Gore") {
-					druid.AddRage(sim, 10.0, rageMetrics)
-					druid.MangleBear.CD.Reset()
+			targetCount := core.TernaryInt32(berserking, 3, 1)
+			numHits := min(targetCount, druid.Env.GetNumTargets())
+			results := make([]*core.SpellResult, numHits)
+
+			for idx := range results {
+				activeStacks := druid.LacerateBleed.Dot(target).GetStacks() + 1
+				activeStacks = core.TernaryInt32(activeStacks > 5, 5, activeStacks)
```
kayla-glick, line 53: `LacerateBleed` has `MaxStacks: 5` so I think thi is redundant?

sim/druid/lacerate.go
```diff
@@ -41,28 +41,33 @@ func (druid *Druid) registerLacerateSpell() {
...
+				baseDamage := spell.Unit.MHWeaponDamage(sim, spell.MeleeAttackPower()) * (.2*float64(activeStacks) + initialDamageMul)
+
+				spell.Cost.FlatModifier -= core.TernaryInt32(berserking, 10, 0)
+				results[idx] = spell.CalcAndDealDamage(sim, target, baseDamage, spell.OutcomeMeleeSpecialHitAndCrit)
+				spell.Cost.FlatModifier += core.TernaryInt32(berserking, 10, 0)
```
kayla-glick, line 58: We can leave this for now but we should really make this cost adjustment inside the berserking aura code

sim/druid/swipe.go
```diff
@@ -102,17 +101,11 @@ func (druid *Druid) newSwipeBearSpellConfig(swipeRank SwipeRankInfo) core.SpellC
 		ThreatMultiplier:             SwipeThreatMultiplier,
 
 		ApplyEffects: func(sim *core.Simulation, target *core.Unit, spell *core.Spell) {
+			damage := baseDamage + .1*druid.GetStat(stats.AttackPower)
```
kayla-glick, line 104: Usually for this you want to use `spell.MeleeAttackPower()`
```suggestion
			damage := baseDamage + .1*spell.MeleeAttackPower()
```

sim/druid/item_sets_pve_phase_4.go
ironclad88: Ah things like the seal of the dawn threat mod are also in this pseudostat, makes sense! I just knew the set bonus was additive with bear threat mod, but it seems to be now if I add some prints to when bear form arua is applied/removed:
Bear form threat bonus before apply 1
Bear form threat bonus after apply 1.65
Bear form threat bonus before remove 1.65
Bear form threat bonus after remove 1

TPS also went up considerably in naxx mode:
Bear form threat bonus before apply 1.8504
Bear form threat bonus after apply 3.05316
Bear form threat bonus before remove 1.65
Bear form threat bonus after remove 1

About 15% higher tps than before this fix. Very important find, thanks!

kayla-glick (approved):

PR #1257, opened 2025-01-28
removed legacy debuffs, added holy sunder

JJwang11:
removed:
degrade (homunculi just sunder now)
crystal yield (effect was incorrect, does not stack with faerie fire)
ancient corrosive poison (from phase 1 no shot this is used at all)

added:
holy sunder (from alliance quest only), actively used


ui/core/components/inputs/buffs_debuffs.ts
```diff
@@ -974,23 +967,13 @@ export const CurseOfVulnerability = makeBooleanDebuffInput({
...
 export const GiftOfArthas = makeBooleanDebuffInput({
-	actionId: player =>
-		player.getMatchingSpellActionId([
-			// SoD Phase 3?
-			{ id: 11374, minLevel: 41 },
-		]),
+	actionId: player => player.getMatchingSpellActionId([{ id: 11374, minLevel: 41 }]),
 	fieldName: 'giftOfArthas',
 });
-export const CrystalYield = makeBooleanDebuffInput({
```
kayla-glick, line 984: If we disable these in the UI then we need to also disable the effects on the back-end because people could have saved data with them enabled still and then won't have any way to turn them off

JJwang11: good catch, i removed the debuffs for applying (and removed the crystal yield/ancient corrosive poison functions for good measure, but maybe we should just leave those in?)

kayla-glick (approved):

PR #1267, opened 2025-01-31
Feature/migrate equip scaling code

1337LutZ:
- Migrates [#1330](https://github.com/wowsims/cata/pull/1330)
- Rewritten Rockbiter to be an Aura that can initialize & be dynamically updated


sim/shaman/enhancement/TestEnhancement.results
1337LutZ, line 1: Minor RNG seed changes.  Values are the same local vs live

PR #1277, opened 2025-02-02
Crit Cap

kayla-glick:


ui/core/player.ts
```diff
@@ -712,18 +715,74 @@ export class Player<SpecType extends Spec> {
...
+					break;
+			}
+		}
+
+		const skillDiff = defenderDefense - weaponSkill;
+		// Due to warrior HS bug, hit cap for crit cap calculation ignores the 19% penalty
+		let meleeHitCap = skillDiff <= 10 ? 5.0 + skillDiff * 0.1 : 5.0 + skillDiff * 0.2 + (skillDiff - 10) * 0.2;
+		meleeHitCap = this.getGear().isDualWielding() && (this.spec !== Spec.SpecWarrior || useDWPenalty) ? meleeHitCap + 19.0 : meleeHitCap + 0.0;
+
+		const dodgeCap = 5.0 + skillDiff * 0.1;
```
sanguinerarogue, line 781: I think these can just have expertise subtracted from them.  That will automatically adjust the expertise cap as well. (just need protection so if you have more expertise then its 0 instead of negative.

PR #1313, opened 2025-02-21
Add some of the new Karazhan random glove suffixes

hillerstorm:
Only Frost Fever, Blood Plague and Obliterate to begin with.
Gotta figure out how Mark of Blood works with % buffs etc before implementing.

Don't know if anything needs to be done for Item Swap since these items are all in-combat stuff 🤔


sim/core/item_swaps.go
```diff
@@ -212,6 +212,40 @@ func (swap *ItemSwap) RegisterActive(itemID int32) {
 	})
 }
 
+// Helper for handling Enchant On Use effects to set a 30s cd on the related spell.
+// Currently only used for random suffix "enchants" like the Karazhan gloves.
+func (swap *ItemSwap) RegisterEnchantActive(effectID int32, spellID int32) {
```
1337LutZ, line 217: Would be nice to merge this with the function above and have a `registerActiveInternal` function that accepts a config to reuse this logic

hillerstorm: There's a few differences though, checking for an item vs checking for an effect but I guess that would be done with the config struct then?

1337LutZ: Yes, similar to the Proc config which can have an ItemID or a SpellID. If both of them are set you could do the `EnchantActive` specific logic

1337LutZ (approved):

PR #1325, opened 2025-03-02
Adding Hunter's Mark as castable spell

ncberman:
Hunters mark overwrites weaker versions of itself
Only 1 hunters mark can be up per hunter


sim/core/sim_concurrent.go
```diff
@@ -498,6 +498,7 @@ func runSimConcurrent(request *proto.RaidSimRequest, progress chan *proto.Progre
 					progress <- msg
 				}
 				log.Printf("Thread %d had an error. Cancelling all sims!", i)
+				log.Printf(msg.FinalRaidResult.Error.Message)
```
kayla-glick, line 501: Did you mean to add this in?

sim/hunter/hunter.go
```diff
@@ -130,6 +131,7 @@ type Hunter struct {
 	CarveMH                 *core.Spell
 	CarveOH                 *core.Spell
 	WingClip                *core.Spell
+	HuntersMark             *core.Spell
```
kayla-glick, line 134: If hunters mark doesn't override in the spellbook with each rank, can we make this an array?

sim/core/sim_concurrent.go
ncberman: I added it because when an error would occur it wouldn't print out any information and just say ""Thread 8 had an error. Cancelling all sims". Not sure if this prints anywhere in the UI I can remove if necessary.

sim/hunter/hunter.go
ncberman: Multiple ranks are available in our spellbook, in the sim I'm also able to select multiple ranks in the sim with this PR. Whats the advantage of making this an array?

kayla-glick: @ncberman it's mostly to have a reference to each spell in case there's ever a reason to use it. If we only keep track of the last one then we lose references to all of the others

sim/hunter/hunters_mark.go
```diff
@@ -0,0 +1,65 @@
...
+			aura.Activate(sim)
+		},
+
+		RelatedAuras: []core.AuraArray{hunter.HuntersMarkAuras},
+	}
+}
+
+func (hunter *Hunter) myHuntersMarkAura(target *core.Unit, playerLevel int32) *core.Aura {
+	return core.HuntersMarkAura(target, hunter.Talents.ImprovedHuntersMark, hunter.Level)
+}
```
kayla-glick, line 53: Hm thinking about this some more, this will make all of the ranks of hunters mark apply the same max available rank aura based on the player's level. Maybe we should just give access to the highest rank spell instead of having all ranks available? Is there ever a reason you'd want to downrank?

kayla-glick (approved):

PR #1330, opened 2025-03-05
BUGFIX - T2 4p / Soul of the Alternator not applying properly

ncberman:
The approach is to apply the damage mod to all shots that aren't the current shot whenever any shot is cast

### DIFF
--- a/sim/hunter/item_sets_pve_phase_5.go
+++ b/sim/hunter/item_sets_pve_phase_5.go
@@ -169,11 +169,7 @@ func (hunter *Hunter) applyT2Ranged4PBonus() {
 		return
 	}
 
-	damageMod := hunter.AddDynamicMod(core.SpellModConfig{
-		Kind:      core.SpellMod_DamageDone_Flat,
-		ClassMask: ClassSpellMask_HunterShots,
-		IntValue:  10,
-	})
+	var damageMod *core.SpellMod
 
 	procAura := hunter.RegisterAura(core.Aura{
 		ActionID: core.ActionID{SpellID: 467312},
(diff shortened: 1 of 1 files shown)


sim/hunter/item_sets_pve_phase_5.go
```diff
@@ -186,19 +182,24 @@ func (hunter *Hunter) applyT2Ranged4PBonus() {
...
 				procAura.Deactivate(sim)
 			}
 
-			hunter.LastShot = spell
+			// Create new damage mod that buffs all shots that are not the current shot
+			damageMod = hunter.AddDynamicMod(core.SpellModConfig{
+				Kind:      core.SpellMod_DamageDone_Flat,
+				ClassMask: ClassSpellMask_HunterShots &^ spell.ClassSpellMask,
+				IntValue:  10,
+			})
```
kayla-glick, line 200: Hm I'm not sure what the performance impact of this could be. Could we potentially just create a map of mods where each key is a shot's class mask, and it corresponds to the mod that applies to all other shots?

ncberman: moved all the possible combinations to map that gets populated OnInit()

sim/hunter/item_sets_pve_phase_5.go
```diff
@@ -169,36 +169,40 @@ func (hunter *Hunter) applyT2Ranged4PBonus() {
...
 		Duration: time.Second * 12,
+		OnInit: func(aura *core.Aura, sim *core.Simulation) {
+			for _, shot := range hunter.Shots {
+				// Create new damage mod that buffs all shots that are not the current shot
+				damageModMap[shot.ClassSpellMask] = hunter.AddDynamicMod(core.SpellModConfig{
+					Kind:      core.SpellMod_DamageDone_Flat,
+					ClassMask: ClassSpellMask_HunterShots &^ shot.ClassSpellMask,
+					IntValue:  10,
+				})
+			}
```
kayla-glick, line 186: I think you can do this outside of the aura

sim/hunter/item_sets_pve_phase_5.go
```diff
@@ -169,36 +169,40 @@ func (hunter *Hunter) applyT2Ranged4PBonus() {
...
 		Handler: func(sim *core.Simulation, spell *core.Spell, _ *core.SpellResult) {
-			if hunter.LastShot != nil && !spell.Matches(hunter.LastShot.ClassSpellMask) {
-				procAura.Activate(sim)
-			} else {
+			if procAura.IsActive() {
 				procAura.Deactivate(sim)
 			}
-
-			hunter.LastShot = spell
+			damageMod = damageModMap[spell.ClassSpellMask]
```
kayla-glick, line 204: You get the damage mod here but you'd need to activate it, too. I think we also want to make sure all others are deactivated maybe? I'm not quite sure how it works

sim/hunter/item_sets_pve_phase_5.go
ncberman: I tried that originally but I don't think the list of hunter shots is populated yet. I could change this to be on spell registered maybe?

sim/hunter/item_sets_pve_phase_5.go
ncberman: The aura activates it on the next line since the variable is declared outside of the proc trigger aura

ncberman: Basically when a shot is cast we deactivate the 'procAura' change the damage mod to apply to the correct shots then reactivate the 'procAura' with the correct shot damage modifier

kayla-glick: Oh derp I see it now

sim/hunter/item_sets_pve_phase_5.go
kayla-glick: `buildMod` uses an `OnSpellRegistered` hook so that should be ok!

kayla-glick (approved):

PR #1335, opened 2025-03-10
BUGFIX - Rogue Murder Talent / Draught of the Sands Interaction

Puddingstomp:
The check for Draught of the Sands is currently contained in an else/if statement after checking for the Murder talent. This means if Murder is selected as a talent then the check for Draught of the Sands will never occur. I moved the check for Draught to its own if statement as the talent and consumable can stack in game and the sim should reflect that.


Puddingstomp:
Just in case it is unclear above, the sim currently allows the user to select both the Murder talent and Draught of the Sands as it should. It just doesn't add the Draught buff to the sim like it should, and it doesn't alert the user anywhere that this interaction is being blocked.

PR #1346, opened 2025-03-16
Free up unused class masks

raethkcj:
And change ClassMasks from int64 to unit64

### DIFF
--- a/sim/shaman/shaman.go
+++ b/sim/shaman/shaman.go
@@ -12,9 +12,9 @@ import (
 var TalentTreeSizes = [3]int{15, 16, 15}
 
 const (
-	ClassSpellMask_ShamanNone int64 = 0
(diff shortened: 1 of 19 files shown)


sim/shaman/shaman.go
```diff
@@ -63,7 +62,7 @@ const (
 	ClassSpellMask_ShamanHealingStreamTotem
 	ClassSpellMask_ShamanManaSpringTotem
 
-	ClassSpellMask_ShamanAll = ClassSpellMask_ShamanLast<<1 - 1
```
raethkcj, line 66: Hmm I hadn't actually checked how `Last` was used here, just assumed it was right before `All`. Doesn't change existing tests but does change the meaning of the mask

raethkcj:
FYI @kayla-glick

PR #1361, opened 2025-03-21
HUNTER - Scarlet Enclave Tier Set Bonuses

ncberman:
Doesn't pass tests because the items don't exist yet to scrape I guess, I tested these implementations by overwriting the crypstalker sets to be dawnstalker.


sim/hunter/item_sets_pve_phase_8.go
```diff
@@ -0,0 +1,216 @@
...
+
+import (
+	"time"
+
+	"github.com/wowsims/sod/sim/core"
+	"github.com/wowsims/sod/sim/core/proto"
+	"github.com/wowsims/sod/sim/core/stats"
+)
+
+var ItemSetDawnstalkerProwess = core.NewItemSet(core.ItemSet{
```
hillerstorm, line 11: The sets don't exist yet in the database so you have to comment out the `core.NewItemSet` calls, both this one and the ranged one (like I did in sim/paladin/item_sets_pve_phase_8.go).
We can uncomment them again after running proper db scrapes

sim/paladin/retribution/TestShockadin1H.results
```diff
@@ -337,6 +349,9 @@ dps_results: {
 dps_results: {
  key: "TestShockadin1H-Phase7-Lvl60-AllItems-EmeraldEncrustedBattleplate"
  value: {
+  dps: 2357.91835
```
hillerstorm, line 352: There seems to have been some merge conflict in this file that also fails the tests. Duplicate dps/tps entries. This whole file should just be reset in this branch :)

hillerstorm (changes requested):

sim/hunter/item_sets_pve_phase_8.go
```diff
@@ -0,0 +1,216 @@
...
+	if hunter.HasAura(label) {
+		return
+	}
+
+	if !hunter.HasRune(proto.HunterRune_RuneBracersRaptorFury) {
+		return
+	}
+
+	hunter.BonusRaptorFuryDamageMultiplier = 0.10
+}
```
kayla-glick, line 97: 🛑 we should still initialize an aura here (just in case). Even though there's no dual application of these sets right now it's good to keep it consistent

sim/hunter/item_sets_pve_phase_8.go
```diff
@@ -0,0 +1,216 @@
...
+		Kind:      core.SpellMod_DamageDone_Flat,
+		ClassMask: ClassSpellMask_HunterStrikes | ClassSpellMask_HunterMongooseBite,
+	})
+
+	core.MakeProcTriggerAura(&hunter.Unit, core.ProcTrigger{
+		Name:           label,
+		ClassSpellMask: ClassSpellMask_HunterStrikes | ClassSpellMask_HunterMongooseBite,
+		Callback:       core.CallbackOnApplyEffects,
+		Handler: func(sim *core.Simulation, spell *core.Spell, result *core.SpellResult) {
+			hasDebuff := result.Target.HasActiveAuraWithTag("SerpentSting") || result.Target.HasActiveAuraWithTag("WyvernStrike - Bleed")
```
kayla-glick, line 46: Checking for an active aura can be pretty expensive so I would probably recommend using a normal aura with an `OnInit` where you can fill an array of spells then loop over them and check `spell.Dot(result.Target).IsActive()`

sim/hunter/item_sets_pve_phase_8.go
```diff
@@ -0,0 +1,218 @@
...
+func (hunter *Hunter) applyScarletEnclaveMelee2PBonus() {
+	label := "S03 - Item - Scarlet Enclave - Hunter - Melee 2P Bonus"
+	if hunter.HasAura(label) {
+		return
+	}
+
+	damageMod := hunter.AddDynamicMod(core.SpellModConfig{
+		Kind:      core.SpellMod_DamageDone_Flat,
+		ClassMask: ClassSpellMask_HunterStrikes | ClassSpellMask_HunterMongooseBite,
+	})
```
kayla-glick, line 40: I actually realized that both this and the T2 Ranged 2p need to be updated. Conditional damage effects use a dummy script that we've found always seems to be equivalent to a % damage multiplier rather than additive. Both should actually use `core.SpellMod_DamageDone_Pct` and then `FloatValue: 1.0`, then in the handler updating the FloatValue conditionally to either 1.20 or 1.0

sim/hunter/item_sets_pve_phase_8.go
kayla-glick: Instead of checking the aura here, what I would recommend doing is change `hunter.SerpentSting` to an array, then use `slices.ContainsFunc` to check if any of the serpent sting spells have a `Dot(result.Target).IsActive()`, and check Wyvern Strike too. Then if either is true apply the 1.20, otherwise 1.0

kayla-glick: The T2 Ranged 2p should also do effectively the same thing with trap. They'd need to be converted to arrays as well, but then you can use the same strategy and it should be a bit more performant

ncberman: so just to understand these types of auras should treated as multiplicative?

Apply Aura: Dummy
Value: 21
PVP Multiplier: 1
Server-side script

PR #1365, opened 2025-03-22
Hunter T4 Bonuses - NEW PR

ncberman:
Not sure what happened to the old PR.. was not being nice to me..

https://www.wowhead.com/classic-ptr/item-set=1936/dawnstalker-prowess
https://www.wowhead.com/classic-ptr/item-set=1937/dawnstalker-armor

### DIFF
--- a/sim/hunter/item_sets_pve_phase_8.go
+++ b/sim/hunter/item_sets_pve_phase_8.go
@@ -1,0 +1,226 @@
+package hunter
+
+import (
+	"time"
+
+	"github.com/wowsims/sod/sim/core"
+	"github.com/wowsims/sod/sim/core/proto"
+	"github.com/wowsims/sod/sim/core/stats"
+)
+
+// TODO: UNCOMMENT WHEN TIER SETS ARE ADDED TO THE SIM
+// var ItemSetDawnstalkerProwess = core.NewItemSet(core.ItemSet{
+// 	Name: "Dawnstalker Prowess",
+// 	Bonuses: map[int32]core.ApplyEffect{
+// 		2: func(agent core.Agent) {
+// 			hunter := agent.(HunterAgent).GetHunter()
+// 			hunter.applyScarletEnclaveMelee2PBonus()
+// 		},
+// 		4: func(agent core.Agent) {
+// 			hunter := agent.(HunterAgent).GetHunter()
+// 			hunter.applyScarletEnclaveMelee4PBonus()
+// 		},
+// 		6: func(agent core.Agent) {
+// 			hunter := agent.(HunterAgent).GetHunter()
+// 			hunter.applyScarletEnclaveMelee6PBonus()
+// 		},
+// 	},
+// })
+
+// Your Strikes and Mongoose Bite deal 20% increased damage to targets afflicted with your Serpent Sting or Wyvern Strike.
+func (hunter *Hunter) applyScarletEnclaveMelee2PBonus() {
+	label := "S03 - Item - Scarlet Enclave - Hunter - Melee 2P Bonus"
+	if hunter.HasAura(label) {
+		return
+	}
+
+	damageMod := hunter.AddDynamicMod(core.SpellModConfig{
+		Kind:      core.SpellMod_DamageDone_Flat,
+		ClassMask: ClassSpellMask_HunterStrikes | ClassSpellMask_HunterMongooseBite,
+	})
+
+	core.MakeProcTriggerAura(&hunter.Unit, core.ProcTrigger{
+		Name:           label,
+		ClassSpellMask: ClassSpellMask_HunterStrikes | ClassSpellMask_HunterMongooseBite,
+		Callback:       core.CallbackOnApplyEffects,
+		Handler: func(sim *core.Simulation, spell *core.Spell, result *core.SpellResult) {
+			hasDebuff := result.Target.HasActiveAuraWithTag("SerpentSting") || result.Target.HasActiveAuraWithTag("WyvernStrike - Bleed")
+			damageMod.UpdateIntValue(core.TernaryInt64(hasDebuff, 20, 0))
+			damageMod.Activate()
+		},
+	})
+}
+
+// Your melee critical strikes increase your attack speed by 20% for 10 sec.
+func (hunter *Hunter) applyScarletEnclaveMelee4PBonus() {
+	label := "S03 - Item - Scarlet Enclave - Hunter - Melee 4P Bonus"
+	if hunter.HasAura(label) {
+		return
+	}
+
+	attackSpeed := 1.20
+
+	procAura := core.MakePermanent(hunter.RegisterAura(core.Aura{
+		Label:    "Wicked Fast",
+		ActionID: core.ActionID{SpellID: 1226357},
+		Duration: time.Second * 10,
+		OnGain: func(_ *core.Aura, sim *core.Simulation) {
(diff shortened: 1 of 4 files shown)


sim/hunter/item_sets_pve_phase_8.go
```diff
@@ -0,0 +1,225 @@
...
+// Increases the bonus damage from Raptor Fury by an additional 10% per stack.
+func (hunter *Hunter) applyScarletEnclaveMelee6PBonus() {
+	label := "S03 - Item - Scarlet Enclave - Hunter - Melee 6P Bonus"
+	if hunter.HasAura(label) {
+		return
+	}
+
+	if !hunter.HasRune(proto.HunterRune_RuneBracersRaptorFury) {
+		return
+	}
```
kayla-glick, line 95: Really small optimization, but can we do this check before the `HasAura` check? It should be less expensive

sim/hunter/item_sets_pve_phase_8.go
```diff
@@ -0,0 +1,225 @@
...
+		Kind:      core.SpellMod_DamageDone_Pct,
+		ClassMask: ClassSpellMask_HunterShots,
+	})
+
+	core.MakeProcTriggerAura(&hunter.Unit, core.ProcTrigger{
+		Name:           label,
+		ClassSpellMask: ClassSpellMask_HunterShots,
+		Callback:       core.CallbackOnApplyEffects,
+		Handler: func(sim *core.Simulation, spell *core.Spell, result *core.SpellResult) {
+			damageMod.UpdateFloatValue(core.TernaryFloat64(hunter.SerpentSting.Dot(result.Target).IsActive(), 1.20, 1.0))
```
kayla-glick, line 141: Do we use multiple ranks of Serpent Sting? It's kind of edge-casey but if someone were to use a lower rank than their max then this wouldn't pick it up

sim/hunter/item_sets_pve_phase_8.go
```diff
@@ -0,0 +1,225 @@
...
+			damageMod.Activate()
+		},
+	})
+}
+
+// Reduces the cooldown on your Chimera Shot, Explosive Shot, and Aimed Shot abilities by 1.5 sec and reduces the cooldown on your Kill Shot ability by 3sec.
```
kayla-glick, line 147: Looks like this comment is copied from T3 and incorrect

sim/hunter/item_sets_pve_phase_8.go
```diff
@@ -0,0 +1,225 @@
...
+
+		OnGain: func(aura *core.Aura, sim *core.Simulation) {
+			hunter.EnableDynamicStatDep(sim, apBonus)
+			hunter.EnableDynamicStatDep(sim, apRangedBonus)
+		},
+		OnExpire: func(aura *core.Aura, sim *core.Simulation) {
+			hunter.DisableDynamicStatDep(sim, apBonus)
+			hunter.DisableDynamicStatDep(sim, apRangedBonus)
+		},
+	})
```
kayla-glick, line 170:
```suggestion
	apBonus := hunter.NewDynamicMultiplyStat(stats.AttackPower, 1.2)
	apRangedBonus := hunter.NewDynamicMultiplyStat(stats.RangedAttackPower, 1.2)

  // Converted to just RegisterAura
	procAura := hunter.RegisterAura(core.Aura{
		Label:    "Wicked Shot",
		ActionID: core.ActionID{SpellID: 1226136},
		Duration: time.Second * 10,
  // Converted to use the new #AttachStatDependency helper
	}).AttachStatDependency(apBonus).AttachStatDependency(apRangedBonus)
```

sim/hunter/item_sets_pve_phase_8.go
kayla-glick: Your ranged critical strikes increase your Attack Power by 20% for 10 sec.

sim/hunter/item_sets_pve_phase_8.go
```diff
@@ -0,0 +1,225 @@
...
+		},
+		OnExpire: func(aura *core.Aura, sim *core.Simulation) {
+			damageMod.Deactivate()
+		},
+		OnCastComplete: func(aura *core.Aura, sim *core.Simulation, spell *core.Spell) {
+			if spell.Matches(ClassSpellMask_HunterMultiShot) {
+				aura.Deactivate(sim)
+			}
+		},
+	})
```
kayla-glick, line 210: We should be able to attach the mod directly to the aura

```suggestion
	multishotAura := hunter.RegisterAura(core.Aura{
		Label:    "Multi-Shot +100% Damage",
		Duration: time.Minute * 5,
		OnCastComplete: func(aura *core.Aura, sim *core.Simulation, spell *core.Spell) {
			if spell.Matches(ClassSpellMask_HunterMultiShot) {
				aura.Deactivate(sim)
			}
		},
	}).AttachSpellMod(core.SpellModConfig{
		Kind:      core.SpellMod_DamageDone_Flat,
		ClassMask: ClassSpellMask_HunterMultiShot,
		IntValue:  100,
	})
```

sim/hunter/item_sets_pve_phase_8.go
```diff
@@ -0,0 +1,225 @@
...
+			hunter.MultiShotBonusTargets = 2
+		},
+	})).AttachProcTrigger(core.ProcTrigger{
+		Name:           label,
+		ClassSpellMask: ClassSpellMask_HunterChimeraShot | ClassSpellMask_HunterKillShot,
+		Callback:       core.CallbackOnApplyEffects,
+		Handler: func(sim *core.Simulation, spell *core.Spell, result *core.SpellResult) {
+			multishotAura.Activate(sim)
+		},
+	})
```
kayla-glick, line 224: Hm for this one you may want to use the `CanProcFromProcs: true` option. I noticed the aura has that flag and I'm guessing it's to allow it to work with the ranged hunter T2.5 kill shots

image

sim/hunter/item_sets_pve_phase_8.go
```diff
@@ -0,0 +1,225 @@
...
+	}
+
+	damageMod := hunter.AddDynamicMod(core.SpellModConfig{
+		Kind:      core.SpellMod_DamageDone_Flat,
+		ClassMask: ClassSpellMask_HunterMultiShot,
+		IntValue:  100,
+	})
+
+	multishotAura := hunter.RegisterAura(core.Aura{
+		Label:    "Multi-Shot +100% Damage",
```
kayla-glick, line 197: Can we add the Action ID here so that it shows up in the UI? https://www.wowhead.com/classic-ptr/spell=1233451/trick-shots

sim/hunter/item_sets_pve_phase_8.go
ncberman: we do not, I can change this by changing the serpent sting field to be a slice of serpent stings in hunter.go if we want to make them all available

sim/hunter/item_sets_pve_phase_8.go
ncberman: I don't think that matters because the T2.5 killshots inherit the class spell mask

sim/hunter/item_sets_pve_phase_8.go
kayla-glick: Oh ok if we don't then that's totally fine

kayla-glick (approved):

Issue #1333, opened 2025-03-07
Enabling/Disabling stats from Stat Weight Computation


balor89:
Computing Stat weights for classes can be compute intensive and demand high iteration count. Currently, for many classes, a large percentage of stat weights computed are either irrelevant (Fire/Nature/Frost Resistance, Stamina, ...) only situationally of interest/relevance (timeworn pieces, sanctified pieces) deterministically known from other stats (agility vs crit, feral attack power vs ap, int vs mana for some classes) or niche and often irrelevant (for feral: int, mp5, spirit, mana). These make stat weight computations take 2-10x as long as they need to for some use cases.

Many of these stats are included because it is desirable for the user to select EPs using them to view gear or because in some scenarios they have value (mana in long fights for feral, for example, but not in short ones) so it's not ideal to just disable all of them entirely. But it would be great to allow to user to check stats to enable or disable during the computation of weights itself.

If this is too much work or infeasible, there could be a list in the backend of stats to be computed that is distinct from stats to be displayed for EPs. That way users can still sort on EPs like Fire Resistance without causing redundant computations for Fire Resistance during stat weight computation time.

PR #1369, opened 2025-03-22
Add ability to disable stats for stat weight calculation

FelixPflaum:
Adds a checkbox column to the table in the stat weight modal to toggle whether stats should be included in the stat weight sim. Feature was requested in https://github.com/wowsims/sod/issues/1333
Disabling them also excludes stats from the EP copy and EP ratios functions. In that case the old values will be kept.

Settings are saved as an independent local storage entry per spec. Don't think it's something you'd want to export and share, so I didn't put it in the sim settings.

grafik

### DIFF
--- a/ui/core/components/stat_weights_action.ts
+++ b/ui/core/components/stat_weights_action.ts
@@ -627,7 +627,7 @@ class EpWeightsMenu extends BaseModal {
 		} else {
 			updateStatEntry(stat.getPseudoStat(), this.swSettings.excludedFromCalc.pseudoStats);
 		}
-		window.localStorage.setItem(this.swSettingsStorageKey, JSON.stringify(this.swSettings).toString());
+		window.localStorage.setItem(this.swSettingsStorageKey, SavedStatWeightSettings.toJsonString(this.swSettings));
 	}
 
 	private setSimProgress(progress: ProgressMetrics) {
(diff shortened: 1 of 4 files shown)


ui/core/components/stat_weights_action.ts
```diff
@@ -542,12 +554,82 @@ class EpWeightsMenu extends BaseModal {
...
+	 */
+	private isEpStat(stat: UnitStat) {
+		if (stat.isStat()) return this.epStats.includes(stat.getStat());
+		return this.epPseudoStats.includes(stat.getPseudoStat());
+	}
+
+	/**
+	 * Check if a stat should be excluded from weight calculation.
+	 * @param stat
+	 * @param pseudoStat
```
kayla-glick, line 593: Looks like this param is missing

ui/core/components/stat_weights_action.ts
```diff
@@ -591,6 +675,20 @@ class EpWeightsMenu extends BaseModal {
...
+			const includeToggleCell = row.querySelector('.swcalc-include-toggle') as HTMLElement;
+			const cb = document.createElement('input');
+			cb.type = 'checkbox';
+			if (stat.isStat() && this.epReferenceStat == stat.getStat()) {
+				cb.checked = true;
+				cb.disabled = true;
+			} else {
+				cb.checked = statIsCalculated;
+				cb.addEventListener('change', () => this.setStatExcluded(stat, !cb.checked));
+			}
```
kayla-glick, line 688: Rather than use an out-of-the-box checkbox could we use a `BooleanPicker` component for this?

ui/core/components/stat_weights_action.ts
FelixPflaum: Yeah I should really start regenerating docs when making changes :D

PR #1401, opened 2025-03-28
P8 Rogue Tank Set Bonus Implementation

fynnen:
Implementing the set bonus https://www.wowhead.com/classic-ptr/item-set=1935/duskwraith-leathers


sim/rogue/item_sets_phase_8.go
```diff
@@ -13,3 +18,126 @@ func (rogue *Rogue) ApplyFallenRegalityRogueBonus(aura *core.Aura) {
...
+
+	rogue.RegisterAura(core.Aura{
+		Label: label,
+		OnInit: func(aura *core.Aura, sim *core.Simulation) {
+			oldOnStacksChange := rogue.RollingWithThePunchesProcAura.OnStacksChange
+			rogue.RollingWithThePunchesProcAura.OnStacksChange = func(aura *core.Aura, sim *core.Simulation, oldStacks int32, newStacks int32) {
+				oldOnStacksChange(aura, sim, oldStacks, newStacks)
+				rogue.PseudoStats.DamageDealtMultiplier /= 1 + 0.01*float64(oldStacks)
+				rogue.PseudoStats.DamageDealtMultiplier *= 1 + 0.01*float64(newStacks)
+			}
```
kayla-glick, line 60: 🔨 Small knit here. We have a helper that simplifies this `Aura#ApplyOnStacksChange`

```suggestion
			rogue.RollingWithThePunchesProcAura.ApplyOnStacksChange(func(aura *core.Aura, sim *core.Simulation, oldStacks int32, newStacks int32) {
				oldOnStacksChange(aura, sim, oldStacks, newStacks)
				rogue.PseudoStats.DamageDealtMultiplier /= 1 + 0.01*float64(oldStacks)
				rogue.PseudoStats.DamageDealtMultiplier *= 1 + 0.01*float64(newStacks)
			})
```

sim/rogue/talents.go
```diff
@@ -440,8 +443,15 @@ func (rogue *Rogue) registerBladeFlurryCD() {
 			// Undo armor reduction to get the raw damage value.
 			curDmg = result.Damage / result.ResistanceMultiplier
 
-			bfHit.Cast(sim, rogue.Env.NextTargetUnit(result.Target))
-			bfHit.SpellMetrics[result.Target.UnitIndex].Casts--
+			curTarget := rogue.Env.NextTargetUnit(result.Target)
+			for i := int32(0); i < rogue.bladeFlurryTargetCount; i++ {
```
kayla-glick, line 447: If we do this without checking the number of targets the user actually has selected it will loop back around to targets you've already hit. I would assign a variable above the aura registration
```
numTargets := rogue.Env.GetNumTargets()
```
then in here, take the min of that and the `bladeFlurryTargetCount`
```
for i := int32(0); i < min(numTargets, rogue.bladeFlurryTargetCount; i++ {
```

sim/rogue/item_sets_phase_8.go
```diff
@@ -13,3 +18,126 @@ func (rogue *Rogue) ApplyFallenRegalityRogueBonus(aura *core.Aura) {
...
+				energyProc.Cast(sim, result.Target)
+			}
+		},
+	})
+
+	core.MakePermanent(rogue.RegisterAura(core.Aura{
+		Label: label,
+		OnInit: func(aura *core.Aura, sim *core.Simulation) {
+			rogue.rollingWithThePunchesBonusHealthStackMultiplier -= 0.02
+			rogue.rollingWithThePunchesMaxStacks += 5
```
kayla-glick, line 130: We can actually do this without using a variable. You should be able to do
```
rogue.RollingWithThePunchesProcAura.MaxStacks += 5
```

sim/rogue/item_sets_phase_8.go
```diff
@@ -13,3 +18,126 @@ func (rogue *Rogue) ApplyFallenRegalityRogueBonus(aura *core.Aura) {
...
+			oldOnStacksChange := rogue.RollingWithThePunchesProcAura.OnStacksChange
+			rogue.RollingWithThePunchesProcAura.OnStacksChange = func(aura *core.Aura, sim *core.Simulation, oldStacks int32, newStacks int32) {
+				oldOnStacksChange(aura, sim, oldStacks, newStacks)
+				if newStacks == 10 {
+					energyAura.Activate(sim)
+				}
+				if newStacks < 10 && oldStacks == 10 {
+					energyAura.Deactivate(sim)
+				}
+			}
```
kayla-glick, line 140: 🔨 We can also use `ApplyOnStacksChange` here, too

sim/rogue/talents.go
fynnen: Oh I was basing the loop on the cleave interaction which has this logic. The loop is broken if we circle back to the original target of the hit
```
				if curTarget == result.Target {
					break
				}
```

sim/rogue/item_sets_phase_8.go
```diff
@@ -13,3 +18,126 @@ func (rogue *Rogue) ApplyFallenRegalityRogueBonus(aura *core.Aura) {
...
+			rogue.rollingWithThePunchesMaxStacks += 5
+			oldOnStacksChange := rogue.RollingWithThePunchesProcAura.OnStacksChange
+			rogue.RollingWithThePunchesProcAura.OnStacksChange = func(aura *core.Aura, sim *core.Simulation, oldStacks int32, newStacks int32) {
+				oldOnStacksChange(aura, sim, oldStacks, newStacks)
+				if newStacks == 10 {
+					energyAura.Activate(sim)
+				}
+				if newStacks < 10 && oldStacks == 10 {
+					energyAura.Deactivate(sim)
+				}
```
kayla-glick, line 139: Another small knit. I think we can safely make this an if-else if

sim/rogue/item_sets_phase_8.go
```diff
@@ -13,3 +18,126 @@ func (rogue *Rogue) ApplyFallenRegalityRogueBonus(aura *core.Aura) {
...
+		ActionID:    core.ActionID{SpellID: 1226956},
+		SpellSchool: core.SpellSchoolNature,
+		ApplyEffects: func(sim *core.Simulation, u *core.Unit, spell *core.Spell) {
+			rogue.AddEnergy(sim, 15, metrics)
+		},
+	})
+
+	energyAura := rogue.RegisterAura(core.Aura{
+		ActionID: core.ActionID{SpellID: 1219291},
+		Label:    "S03 - Item - Scarlet Enclave - Rogue - Tank 6P Bonus Energy Gain",
```
kayla-glick, line 118: Looks like this aura is missing a `Duration` so it should error out. I think you can use `Duration: core.NeverExpires` since it's dynamically enabled/disabled by the RWTP aura stacks changing

sim/rogue/item_sets_phase_8.go
```diff
@@ -13,3 +18,126 @@ func (rogue *Rogue) ApplyFallenRegalityRogueBonus(aura *core.Aura) {
...
+
+	if rogue.HasAura(label) {
+		return
+	}
+
+	metrics := rogue.NewEnergyMetrics(core.ActionID{SpellID: 1226956})
+
+	energyProc := rogue.RegisterSpell(core.SpellConfig{
+		ActionID:    core.ActionID{SpellID: 1226956},
+		SpellSchool: core.SpellSchoolNature,
```
kayla-glick, line 110: You should be able to skip a spell school here, but we should add some flags to avoid callbacks:
```
Flags: core.NoOnCastComplete
```

sim/rogue/runes.go
```diff
@@ -361,17 +361,18 @@ func (rogue *Rogue) applyRollingWithThePunches() {
...
 	}
 
-	statDeps := make([]*stats.StatDependency, 6) // 5 stacks + zero condition
-	for i := 1; i < 6; i++ {
-		statDeps[i] = rogue.NewDynamicMultiplyStat(stats.Health, 1.0+.06*float64(i))
-
+	rogue.rollingWithThePunchesBonusHealthStackMultiplier += 0.06
+	rogue.rollingWithThePunchesMaxStacks += 5
+	statDeps := make([]*stats.StatDependency, rogue.rollingWithThePunchesMaxStacks+1) // 5 stacks + zero condition
+	for i := 1; i < rogue.rollingWithThePunchesMaxStacks+1; i++ {
```
fynnen, line 367: I think this does not work with the property incremented in phase 8 6p aura because it's created at `applyRWTPRune` and never recalculated. Was it done to minimize compute efforts in OnStackChanges @kayla-glick ? I think to make it truly dynamic with the 6p aura, I need to move the 
`rogue.NewDynamicMultiplyStat(stats.Health, 1.0+rogue.rollingWithThePunchesBonusHealthStackMultiplier*float64(i))` 
inside the aura like this

`aura.Unit.DisableDynamicStatDep(sim, rogue.NewDynamicMultiplyStat(stats.Health, 1.0+rogue.rollingWithThePunchesBonusHealthStackMultiplier*float64(oldStacks)))`

fynnen: I don't know if NewDynamicMultiplyStat is costly to call on every stack change

sim/rogue/item_sets_phase_8.go
fynnen: Just to make sure, I don't need to call `oldOnStacksChange` with this helper right? It chains whatever callback is sent to `ApplyOnStacksChange`?

kayla-glick: Yep it handles it all within the helper!

sim/rogue/talents.go
kayla-glick: Oh that works too! It's probably not quite as graceful but if you just copied another example that's ok too

sim/rogue/runes.go
kayla-glick: @fynnen oh yeah you're right. You would have to increment the attributes outside of the aura in the set code, which is totally fine to do in this case. Usually as long as you're not trying to reference a spell or aura in the to-level code because chances are they don't exist yet.

> I don't know if NewDynamicMultiplyStat is costly to call on every stack change
You can't create stat dependencies once the sim has gotten to that point. If you tried it would do a `panic`

sim/rogue/talents.go
kayla-glick: Oh looking at the Cleave code, I'm pretty sure I was trying to refactor it but then forgot to delete the target comparison

sim/rogue/runes.go
fynnen: I might not be understanding the lifecycle correctly, but is `applyRollingWithThePunches` called after all auras are set? Because if it's done _before_, `statDeps` will never be recalculated right?

sim/rogue/item_sets_phase_8.go
fynnen: Yeah I reverted this change because it wasn't working, but adding to the maxStacks property of the aura doesn't seem to be working either when I run my sim. I still only get up to 5 stacks of the aura.

sim/rogue/runes.go
kayla-glick: @fynnen this line setting up the stat dependencies is called before auras are initialized. That's why we'd have to move the set bonus increment the variable to outside of the aura

sim/rogue/item_sets_phase_8.go
kayla-glick: Oh, hm ok. I wonder why 🤔 I took a similar approach in the Shamans 6pT4

sim/rogue/runes.go
fynnen: It worked by moving the set bonus increments outside of the aura! Thanks!

sim/rogue/item_sets_phase_8.go
fynnen: I think it was my browser cache messing something 😭 It works fine now

sim/rogue/item_sets_phase_5.go
```diff
@@ -165,11 +165,11 @@ func (rogue *Rogue) applyT2Tank4PBonus() {
...
-			oldOnStacksChange := rogue.RollingWithThePunchesProcAura.OnStacksChange
-			rogue.RollingWithThePunchesProcAura.OnStacksChange = func(aura *core.Aura, sim *core.Simulation, oldStacks int32, newStacks int32) {
-				oldOnStacksChange(aura, sim, oldStacks, newStacks)
-				rogue.AddStatDynamic(sim, stats.Armor, float64(0.2*initarmor*float64(newStacks-oldStacks)))
-			}
+			rogue.RollingWithThePunchesProcAura.ApplyOnStacksChange(func(aura *core.Aura, sim *core.Simulation, oldStacks int32, newStacks int32) {
+				if newStacks <= 5 && oldStacks <= 5 {
+					rogue.AddStatDynamic(sim, stats.Armor, float64(0.2*initarmor*float64(newStacks-oldStacks)))
+				}
+			})
```
kayla-glick, line 172: Hm I'm not sure this quite works? If you're at 100% additional armor and then go from 10 stacks => 0 it should remove 5 stacks worth of armor, but because of the condition it won't

sim/rogue/item_sets_phase_8.go
```diff
@@ -13,3 +18,123 @@ func (rogue *Rogue) ApplyFallenRegalityRogueBonus(aura *core.Aura) {
...
+	}
+
+	rogue.rollingWithThePunchesBonusHealthStackMultiplier -= 0.02
+	rogue.rollingWithThePunchesMaxStacks += 5
+
+	metrics := rogue.NewEnergyMetrics(core.ActionID{SpellID: 1226957})
+
+	energyProc := rogue.RegisterSpell(core.SpellConfig{
+		ActionID: core.ActionID{SpellID: 1226957},
+		Flags:    core.SpellFlagNoOnCastComplete,
```
kayla-glick, line 111: Small knit but I think we could use the `core.SpellFlagNoLifecycleCallbacks` flag here instead to save just a little bit of performance. That will stop this spell from triggering `OnApplyEffects` callbacks, which should be fine since there' not really anything that messes with what it does

fynnen: Oh sure! Done

kayla-glick (approved):

PR #1414, opened 2025-03-29
Regicide Weapon for Rogue/Warrior/Hunter

ncberman:
https://www.wowhead.com/classic-ptr/item=240851/regicide?spellModifier=446658


sim/rogue/items.go
```diff
@@ -225,3 +225,50 @@ func (rogue *Rogue) ApplyQueensfallRogueEffect(aura *core.Aura) {
 		},
 	})
 }
+
+// Striking a higher level enemy applies a stack of Coup, increasing their damage taken from your next Envenom by 5% per stack, stacking up to 20 times.
+func (rogue *Rogue) ApplyRegicideRogueEffect(aura *core.Aura) {
+	// Coup debuff array
+	debuffAuras := rogue.NewEnemyAuraArray(func(unit *core.Unit, _ int32) *core.Aura {
+		aura := unit.RegisterAura(core.Aura{
+			ActionID:  core.ActionID{SpellID: 1231417},
```
kayla-glick, line 234: I found the debuff the game uses https://www.wowhead.com/classic-ptr/spell=1231424/coup it has a duration of 15s

sim/hunter/items.go
```diff
@@ -643,3 +643,50 @@ func (hunter *Hunter) newBloodlashProcItem(bonusStrength float64, spellID int32)
 
 	hunter.ItemSwap.RegisterProc(BloodlashBow, aura)
 }
+
+// Striking a higher level enemy applies a stack of Coup, increasing their damage taken from your next Kill Shot by 5% per stack, stacking up to 20 times.
+func (hunter *Hunter) ApplyRegicideHunterEffect(aura *core.Aura) {
+	// Coup debuff array
+	debuffAuras := hunter.NewEnemyAuraArray(func(unit *core.Unit, _ int32) *core.Aura {
+		aura := unit.RegisterAura(core.Aura{
+			ActionID:  core.ActionID{SpellID: 1231417},
```
kayla-glick, line 652: Here's the debuff used for hunter https://www.wowhead.com/classic-ptr/spell=1231765/coup also a 15s duration

sim/warrior/items.go
```diff
@@ -253,3 +253,50 @@ func (warrior *Warrior) ApplyQueensfallWarriorEffect(aura *core.Aura) {
 		},
 	})
 }
+
+// Striking a higher level enemy applies a stack of Coup, increasing their damage taken from your next Execute by 10% per stack, stacking up to 20 times. At 20 stacks, Execute may be cast regardless of the target's health.
+func (warrior *Warrior) ApplyRegicideWarriorEffect(aura *core.Aura) {
+	// Coup debuff array
+	debuffAuras := warrior.NewEnemyAuraArray(func(unit *core.Unit, _ int32) *core.Aura {
+		aura := unit.RegisterAura(core.Aura{
+			ActionID:  core.ActionID{SpellID: 1231417},
```
kayla-glick, line 262: And here's the warrior aura https://www.wowhead.com/classic-ptr/spell=1231416/coup same deal as the others!

sim/warrior/execute.go
```diff
@@ -48,7 +48,7 @@ func (warrior *Warrior) registerExecuteSpell() {
 			},
 		},
 		ExtraCastCondition: func(sim *core.Simulation, target *core.Unit) bool {
-			return sim.IsExecutePhase20() || (hasSuddenDeathRune && warrior.SuddenDeathAura.IsActive())
+			return sim.IsExecutePhase20() || (hasSuddenDeathRune && warrior.SuddenDeathAura.IsActive()) || (target.HasActiveAura("Coup") && target.GetAura("Coup").GetStacks() == 20)
```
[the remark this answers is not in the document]
kayla-glick: It could be useful to create a `Spell#ApplyExtraCastCondition` method similar to the Aura "Apply X Callback" methods to simplify the process

sim/warrior/items.go
```diff
@@ -253,3 +253,50 @@ func (warrior *Warrior) ApplyQueensfallWarriorEffect(aura *core.Aura) {
...
+	aura.AttachProcTrigger(core.ProcTrigger{
+		Name:     "Regicide Trigger - Warrior",
+		Callback: core.CallbackOnSpellHitDealt,
+		Outcome:  core.OutcomeLanded,
+		ProcMask: core.ProcMaskMelee,
+		Handler: func(sim *core.Simulation, spell *core.Spell, result *core.SpellResult) {
+			debuffAuras[result.Target.Index].Activate(sim)
+			debuffAuras[result.Target.Index].AddStack(sim)
+			exeDamageMod.UpdateFloatValue(1 + float64(debuffAuras[result.Target.Index].GetStacks())*0.10)
+			exeDamageMod.Activate()
```
kayla-glick, line 299: Since all of the coup debuffs are on a single target and only affect damage done to that target, we would need to apply this a bit differently, as in a multi target sim this won't respect target swapping.

I think what we can do instead is add a `OnApplyEffects` where we check if the spell is execute, then update and activate the damage mod, rather than setting it here. That should apply to all of the verions of coup

sim/warrior/execute.go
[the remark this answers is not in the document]
kayla-glick: @ncberman looks like this got left in stitll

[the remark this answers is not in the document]
ncberman: should be good now

sim/warrior/items.go
ncberman: Created a new proc trigger for OnApplyEffects that updates the damage mod when envenom/execute/killshot is cast based on the debuff stacks of the enemy. Tested this works by having a sim of warrior that would try to execute target 2 while only hitting target 1.

PR #1421, opened 2025-03-30
P8 rogue dps tier

Puddingstomp:
This is a sanity check draft PR for me to make sure I haven't gone completely down the wrong path with my implementation so far with the Phase 8 Rogue DPS Tier set bonuses. I'm very new to Go and also very new to contributing to a community project. Appreciate any feedback anyone has.


Puddingstomp:
For anyone looking, I think 2pc and 6pc might be complete-ish, 4pc is still a work in progress and not really close to implemented.

sim/rogue/rogue.go
```diff
@@ -157,6 +157,10 @@ type Rogue struct {
 	sebaciousPoisonDebuffAura core.AuraArray
 	atrophicPoisonDebuffAura  core.AuraArray
 	numbingPoisonDebuffAura   core.AuraArray
+
+	// working on p8 dps tier
+	BleedsActive  int
+	PoisonsActive int
```
kayla-glick, line 163: This `BleedsActive` approach actually doesn't work correctly in multi-target sims I just realized 😅 the Feral code will need to be fixed up, too. I think there's a way we can do it though

The problem is that these effects need to know how many bleeds/poisons are active on a given target, not just cumulatively. To do that we probably need to make this an array consisting of all of the enemy unit indexes with a count for each.

The data type we can probably use here is a `map[int]int` for quick and efficient lookup
```
BleedsActive map[int]int
PoisonsActive map[int]int
```

Then in the `OnReset` function we would want to set the values of all of the enemy unit indexes to a count of 0:

```
for _, target := range sim.Encounter.TargetUnits {
  rogue.BleedsActive[target.UnitIndex] = 0
  rogue.PoionsActive[target.UnitIndex] = 0
}
```

sim/rogue/rupture.go
```diff
@@ -48,6 +48,12 @@ func (rogue *Rogue) registerRupture() {
 			Aura: core.Aura{
 				Label: "Rupture",
 				Tag:   RogueBleedTag,
+				OnGain: func(aura *core.Aura, sim *core.Simulation) {
+					rogue.BleedsActive++
+				},
```
kayla-glick, line 53: With the above changes, these `OnGain`s would look like
```suggestion
				OnGain: func(aura *core.Aura, sim *core.Simulation) {
					rogue.BleedsActive[aura.Unit.TargetIndex]++
				},
```

sim/rogue/rupture.go
```diff
@@ -48,6 +48,12 @@ func (rogue *Rogue) registerRupture() {
...
+				OnExpire: func(aura *core.Aura, sim *core.Simulation) {
+					rogue.BleedsActive--
+				},
```
kayla-glick, line 56: And the `OnExpire`s would look like
```suggestion
				OnExpire: func(aura *core.Aura, sim *core.Simulation) {
					rogue.BleedsActive[aura.Unit.TargetIndex]--
				},
```

sim/rogue/item_sets_phase_8.go
```diff
@@ -1,6 +1,113 @@
...
+		ClassMask:  ClassSpellMask_RogueBackstab | ClassSpellMask_RogueSinisterStrike | ClassSpellMask_RogueSaberSlash | ClassSpellMask_RogueMutilate,
+		FloatValue: 1.0,
+	})
+
+	core.MakeProcTriggerAura(&rogue.Unit, core.ProcTrigger{
+		ActionID:       core.ActionID{SpellID: 1226843}, // Tracking in APL
+		Name:           label,
+		Callback:       core.CallbackOnApplyEffects,
+		ProcChance:     0.1,
+		ClassSpellMask: ClassSpellMask_RogueBackstab | ClassSpellMask_RogueSinisterStrike | ClassSpellMask_RogueSaberSlash | ClassSpellMask_RogueMutilate,
```
kayla-glick, line 57: Just to reduce the risk of an error, can we assign the class mask to a single variable and pass it to both places?

sim/rogue/item_sets_phase_8.go
```diff
@@ -1,6 +1,113 @@
...
+			fmt.Println("Bleeds: ", rogue.BleedsActive)
+			fmt.Println("Poisons: ", rogue.PoisonsActive)
+
+			// Only apply the damage mod up to 3 times for the 30% bonus maximum
+			fmt.Println("2P proc triggered")
+			if totalBleedsAndPoisons <= 3 {
+				fmt.Println("Applying 2P damage mod, value: ", totalBleedsAndPoisons)
+				damageMod.Activate()
+				damageMod.UpdateFloatValue(1 + 0.10*float64(totalBleedsAndPoisons))
+			}
```
kayla-glick, line 68:
```suggestion
			damageMod.UpdateFloatValue(1 + 0.10*float64(min(3, totalBleedsAndPoisons)))
			damageMod.Activate()
```

sim/rogue/item_sets_phase_8.go
```diff
@@ -1,6 +1,113 @@
...
+// Your Poison and autoattack critical strikes have a 10% chance to grant you a combo point. (Proc chance: 10%)
+func (rogue *Rogue) applyScarletEnclaveDamage4PBonus() {
+	label := "S03 - Item - Scarlet Enclave - Rogue - Damage 4P Bonus"
+	if rogue.HasAura(label) {
+		return
+	}
+
+	comboPointMetrics := rogue.NewComboPointMetrics(core.ActionID{SpellID: 1226869})
+
+	// TODO: Figure out how to add autoattacks to the spell matches
```
kayla-glick, line 82: To do this you basically just want to create a second proc trigger or do a custom aura. Which ever way is totally fine really. if you go for a second trigger, just skipp the `ClassSpellMask` on that one

sim/rogue/item_sets_phase_8.go
```diff
@@ -1,6 +1,113 @@
...
+	core.MakeProcTriggerAura(&rogue.Unit, core.ProcTrigger{
+		Name:           label,
+		ClassSpellMask: ClassSpellMask_RogueDeadlyPoisonTick | ClassSpellMask_RogueOccultPoisonTick | ClassSpellMask_RogueInstantPoison,
+		Callback:       core.CallbackOnPeriodicDamageDealt | core.CallbackOnSpellHitDealt,
+		Outcome:        core.OutcomeLanded,
+		Handler: func(sim *core.Simulation, spell *core.Spell, result *core.SpellResult) {
+			if spell.Matches(ClassSpellMask_RogueDeadlyPoisonTick|ClassSpellMask_RogueOccultPoisonTick|ClassSpellMask_RogueInstantPoison) && result.DidCrit() && sim.Proc(0.10, "Combo!") {
```
kayla-glick, line 89: You can skip checking the class mask here because you already applied a filter to the proc trigger aura that does the same thing

sim/rogue/item_sets_phase_8.go
```diff
@@ -1,6 +1,113 @@
...
+	}
+
+	comboPointMetrics := rogue.NewComboPointMetrics(core.ActionID{SpellID: 1226869})
+
+	// TODO: Figure out how to add autoattacks to the spell matches
+	core.MakeProcTriggerAura(&rogue.Unit, core.ProcTrigger{
+		Name:           label,
+		ClassSpellMask: ClassSpellMask_RogueDeadlyPoisonTick | ClassSpellMask_RogueOccultPoisonTick | ClassSpellMask_RogueInstantPoison,
+		Callback:       core.CallbackOnPeriodicDamageDealt | core.CallbackOnSpellHitDealt,
+		Outcome:        core.OutcomeLanded,
```
kayla-glick, line 87: Looks like you want this to be on crits, so you can actually do
```suggestion
		Outcome:        core.OutcomeCrit,
```

And then remove it from inside the handler

sim/rogue/item_sets_phase_8.go
kayla-glick, line 89: And then for the proc chance, you can actually pass another option to the Proc Trigger config:
```
ProcChance: 0.10
```
and then remove this here.

Then you handler just becomes the line adding combo points

sim/rogue/item_sets_phase_8.go
```diff
@@ -1,6 +1,113 @@
...
+	}
+
+	core.MakePermanent(rogue.RegisterAura(core.Aura{
+		Label: label,
+	}).AttachSpellMod(core.SpellModConfig{
+		Kind:      core.SpellMod_DamageDone_Flat,
+		ClassMask: ClassSpellMask_RogueAmbush | ClassSpellMask_RogueEviscerate | ClassSpellMask_RogueCrimsonTempest | ClassSpellMask_RogueEnvenom,
+		IntValue:  50,
+	}))
+}
```
kayla-glick, line 110: Perfect 👍 ✨

sim/rogue/poisons.go
```diff
@@ -148,6 +152,10 @@ func (rogue *Rogue) applyDeadlyBrewDeadly() {
 				rogue.DeadlyPoison[DeadlyBrewProc].Cast(sim, result.Target)
 			}
 		},
+		OnExpire: func(aura *core.Aura, sim *core.Simulation) {
+			fmt.Println("Deadly Brew (Deadly) deactivated")
+			rogue.PoisonsActive--
+		},
```
kayla-glick, line 158: This aura is a bit missleading but it's actually just acting as a proc trigger for deadly brew, not an actual poison effect

sim/rogue/poisons.go
```diff
@@ -199,9 +210,14 @@ func (rogue *Rogue) applyDeadlyPoison() {
 				rogue.DeadlyPoison[NormalProc].Cast(sim, result.Target)
 			}
 		},
+		OnExpire: func(aura *core.Aura, sim *core.Simulation) {
+			fmt.Println("Deadly Poison deactivated")
+			rogue.PoisonsActive--
+		},
```
kayla-glick, line 216: Same as above here

sim/rogue/poisons.go
```diff
@@ -224,9 +242,14 @@ func (rogue *Rogue) applyWoundPoison() {
 				rogue.WoundPoison[NormalProc].Cast(sim, result.Target)
 			}
 		},
+		OnExpire: func(aura *core.Aura, sim *core.Simulation) {
+			fmt.Println("Wound Poison deactivated")
+			rogue.PoisonsActive--
+		},
```
kayla-glick, line 248: Same as above here

sim/rogue/poisons.go
```diff
@@ -250,9 +275,14 @@ func (rogue *Rogue) applyOccultPoison() {
 				rogue.OccultPoison[NormalProc].Cast(sim, result.Target)
 			}
 		},
+		OnExpire: func(aura *core.Aura, sim *core.Simulation) {
+			fmt.Println("Occult Poison deactivated")
+			rogue.PoisonsActive--
+		},
```
kayla-glick, line 281: Same as above here

sim/rogue/poisons.go
```diff
@@ -274,9 +306,14 @@ func (rogue *Rogue) applySebaciousPoison() {
 				rogue.SebaciousPoison[NormalProc].Cast(sim, result.Target)
 			}
 		},
+		OnExpire: func(aura *core.Aura, sim *core.Simulation) {
+			fmt.Println("Sebacious Poison deactivated")
+			rogue.PoisonsActive--
+		},
```
kayla-glick, line 312: Same as above here

sim/rogue/poisons.go
```diff
@@ -298,9 +337,14 @@ func (rogue *Rogue) applyAtrophicPoison() {
 				rogue.AtrophicPoison[NormalProc].Cast(sim, result.Target)
 			}
 		},
+		OnExpire: func(aura *core.Aura, sim *core.Simulation) {
+			fmt.Println("Atrophic Poison deactivated")
+			rogue.PoisonsActive--
+		},
```
kayla-glick, line 343: Ssame as above here

sim/rogue/poisons.go
```diff
@@ -322,6 +368,10 @@ func (rogue *Rogue) applyNumbingPoison() {
 				rogue.NumbingPoison[NormalProc].Cast(sim, result.Target)
 			}
 		},
+		OnExpire: func(aura *core.Aura, sim *core.Simulation) {
+			fmt.Println("Numbing Poison deactivated")
+			rogue.PoisonsActive--
+		},
```
kayla-glick, line 374: Same as above here

sim/rogue/poisons.go
```diff
@@ -322,6 +368,10 @@ func (rogue *Rogue) applyNumbingPoison() {
...
 	})
```
kayla-glick, line 375: What you're looking for are the functions `func (rogue *Rogue) makeXPoison(procSource ProcPoisonSource)`. Those are the ones actually building the spells that act as the poisons / debuff applicators

kayla-glick: For example, here is where you probably want to add `OnGain` and `OnExpire` to increment for Occult
image

sim/rogue/item_sets_phase_8.go
```diff
@@ -1,6 +1,121 @@
...
+	// Add a combo point for autoattack crit | Not sure if I need to add anything here to track autoattacks
+	core.MakeProcTriggerAura(&rogue.Unit, core.ProcTrigger{
+		Name:       label,
+		Callback:   core.CallbackOnPeriodicDamageDealt | core.CallbackOnSpellHitDealt,
+		Outcome:    core.OutcomeCrit,
+		ProcChance: 0.10,
+		Handler: func(sim *core.Simulation, spell *core.Spell, result *core.SpellResult) {
+			rogue.AddComboPoints(sim, 1, rogue.CurrentTarget, comboPointMetrics)
+		},
+	})
```
kayla-glick, line 100: On this proc trigger, in order to make it only proc from auto attacks and not from other things like melee abilities, spell hits, etc, you would want to set the `ProcMask` option to restrict the ProcMasks that can proc it
```
ProcMask: core.ProcMaskMeleeWhiteHit
```

Puddingstomp:
I am a little unsure of the best way to handle the issue I have run into with Sebacious Poison, Atrophic Poison, and Numbing Poison. I have added OnGain and OnExpire functions to the necessary poisons that have their Aura constructor inside of poisons.go but it looks like the Aura constructors for these 3 poisons are inside debuffs.go and have some other functions that rely on them there, and then in poisons.go we are just returning the Aura created in debuffs.go.

My other issue here is adding this stuff to debuffs.go. That doesn't really seem like a good idea to me because I am just adding a new set bonus for rogue, it seems maybe not good to do it that way? I also don't have access to the necessary Rogue struct in debuffs.go. A lot of my "issues" are probably me not understanding the Go language very well but any advice for how to do this properly would be appreciated.

kayla-glick:
Ah that makes sense @Puddingstomp. What you could do is once you initialize the debuff array in each spot, loop over the debuffs and add an `ApplyOnGain` and `ApplyOnExpire` that do what you're looking for

Puddingstomp:
Wow that helped immensely. I think I understand now what you mean by loop over the debuffs because at this point I have added poison tracking OnGain and OnExpire functions to the Register functions for 6 different poisons. Would it be ok to write a new helper function in poisons.go specifically for the set bonus that will add those ApplyOnGain and ApplyOnExpire functions to all of the necessary poisons?

kayla-glick:
@Puddingstomp absolutely! I think it makes sense to simplify it into one spot

Puddingstomp:
I'm still struggling to get a helper function to work at all. I think it's still scope related. It seems like I can either get it working for the poisons where the Aura constructor lives in poisons.go but not the 3 where the constructors live in debuffs.go or the other way around. I've not been able to get it to work for both.

I will push where I'm at shortly but I think I just need help consolidating all the OnGains and OnExpires into a single helper function and getting the scopes to cooperate.

Puddingstomp:
This all appears to be working from my initial tests. I don't know the best way to log when the 4pc bonus is occurring and the Combo! spell (id=1226868) isn't showing in the Results tab anywhere that I saw. That may be because of the spell id I'm using in item_sets_phase_8.go on line 74, maybe that should be the spell id for Combo! instead?

sim/rogue/rogue.go
```diff
@@ -183,6 +187,16 @@ func (rogue *Rogue) builderFlags() core.SpellFlag {
...
 func (rogue *Rogue) Initialize() {
+	// p8 DPS tier bonus tracking
+	rogue.BleedsActive = make(map[int32]int32, len(rogue.Env.Encounter.TargetUnits))
+	for _, target := range rogue.Env.Encounter.TargetUnits {
+		rogue.BleedsActive[target.UnitIndex] = 0
+	}
+	rogue.PoisonsActive = make(map[int32]int32, len(rogue.Env.Encounter.TargetUnits))
+	for _, target := range rogue.Env.Encounter.TargetUnits {
+		rogue.PoisonsActive[target.UnitIndex] = 0
+	}
```
kayla-glick, line 198: 🔨 Small knit here, but this could just be one loop where you assign both! You'd still need both `make`s of course

sim/rogue/poisons.go
```diff
@@ -67,6 +67,16 @@ func (rogue *Rogue) improvedPoisonsBonusProcChance() float64 {
 	return 0.02 * float64(rogue.Talents.ImprovedPoisons)
 }
 
+// p8 DPS tier bonus helper function
+func trackTotalUniquePoisons(aura *core.Aura, rogue *Rogue) {
+	aura.ApplyOnGain(func(aura *core.Aura, sim *core.Simulation) {
+		rogue.PoisonsActive[aura.Unit.UnitIndex]++
+	})
+	aura.ApplyOnExpire(func(aura *core.Aura, sim *core.Simulation) {
```
kayla-glick, line 75: 🔨 Tiny knit. We can chain these calls because they return the Aura pointer
```suggestion
	}).ApplyOnExpire(func(aura *core.Aura, sim *core.Simulation) {
```

sim/rogue/item_sets_phase_8.go
```diff
@@ -7,18 +7,114 @@ import (
...
+	// Add a combo point for autoattack crit
+	core.MakeProcTriggerAura(&rogue.Unit, core.ProcTrigger{
+		Name:       label,
+		Callback:   core.CallbackOnPeriodicDamageDealt | core.CallbackOnSpellHitDealt,
+		Outcome:    core.OutcomeCrit,
+		ProcChance: 0.10,
+		ProcMask:   core.ProcMaskMeleeWhiteHit,
+		Handler: func(sim *core.Simulation, spell *core.Spell, result *core.SpellResult) {
+			rogue.AddComboPoints(sim, 1, rogue.CurrentTarget, comboPointMetrics)
+		},
```
kayla-glick, line 96: Oh I just realized these both share the same `Name`. I'm surprised that that isn't blowing up 🤔. Maybe it would be worth combining them into one custom aura

```suggestion
  procClassMasks := ClassSpellMask_RogueDeadlyPoisonTick | ClassSpellMask_RogueOccultPoisonTick | ClassSpellMask_RogueInstantPoison
  
  core.MakePermanent(rogue.RegisterAura(core.Aura{
    Label: label
    OnPeriodicDamageDealt: func(aura *core.Aura, sim *core.Simulation, spell *core.Spell, result *core.SpellResult) {
      if result.DidCrit() && spell.Matches(procClassMasks) && sim.Proc(0.10, "Combo! proc") {
        rogue.AddComboPoints(sim, 1, rogue.CurrentTarget, comboPointMetrics)
      }
    },
    OnSpellHitDealt: func(aura *core.Aura, sim *core.Simulation, spell *core.Spell, result *core.SpellResult) {
      if result.DidCrit() && (spell.Matches(procClassMasks) || spell.ProcMask.Matches(core.ProcMaskMeleeWhiteHit) && sim.Proc(0.10, "Combo! proc") {
        rogue.AddComboPoints(sim, 1, rogue.CurrentTarget, comboPointMetrics)
      }
    }
  }))
```

sim/rogue/rogue.go
Puddingstomp: much cleaner

sim/rogue/item_sets_phase_8.go
```diff
@@ -7,16 +7,104 @@ import (
...
+		return
+	}
+
+	// Combo! spell that adds the combo point for the set bonus
+	comboPointMetrics := rogue.NewComboPointMetrics(core.ActionID{SpellID: 1226869})
+
+	procClassMasks := ClassSpellMask_RogueDeadlyPoisonTick | ClassSpellMask_RogueOccultPoisonTick | ClassSpellMask_RogueInstantPoison
+
+	core.MakePermanent(rogue.RegisterAura(core.Aura{
+		Label: label,
```
kayla-glick, line 78: Sorry I thought you were adding the ActionID here or I would've mentioned it sooner! To expose this as the 6pc aura we'd want to add
```
ActionID: core.ActionID{SpellID: 1226869}
```

sim/rogue/item_sets_phase_8.go
```diff
@@ -7,16 +7,104 @@ import (
...
+// 4P
+// Your Poison and autoattack critical strikes have a 10% chance to grant you a combo point. (Proc chance: 10%)
+func (rogue *Rogue) applyScarletEnclaveDamage4PBonus() {
+	label := "S03 - Item - Scarlet Enclave - Rogue - Damage 4P Bonus"
+	if rogue.HasAura(label) {
+		return
+	}
+
+	// Combo! spell that adds the combo point for the set bonus
+	comboPointMetrics := rogue.NewComboPointMetrics(core.ActionID{SpellID: 1226869})
```
kayla-glick, line 73: Also in hindsight using the 4pc spell ID like you did here is good 👍 Even though the spell is technically the `Combo!` one, I think it will be clearer for players the way you did it here

kayla-glick (approved):

PR #1427, opened 2025-03-31
P8 Rogue tank t4 set bonus updates

fynnen:
changing how the energy proc works from only dodges to dodges and parries, and now 10 energy.


fynnen:
image

kayla-glick (approved):

PR #1428, opened 2025-04-01
Implement the 4p and 6p set bonus of guardian druid t4

fynnen:
https://www.wowhead.com/classic-ptr/item-set=1948/waywatcher-guardian

4p: increase berserk aura duration by 15 seconds
6p: triggers a 30% haste buff that lasts 10 seconds every time you crit

Not including the 2p since it is only mitigation, and need to look into how to implement it.

### DIFF
--- a/sim/druid/item_sets_pve_phase_8.go
+++ b/sim/druid/item_sets_pve_phase_8.go
@@ -344,6 +344,9 @@ func (druid *Druid) applyScarletEnclaveGuardian4PBonus() {
 
 	core.MakePermanent(druid.RegisterAura(core.Aura{
 		Label: label,
+		OnInit: func(aura *core.Aura, sim *core.Simulation) {
+			druid.BerserkAura.Duration += time.Second * 15
+		},
 	}))
 }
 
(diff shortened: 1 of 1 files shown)


sim/druid/item_sets_pve_phase_8.go
```diff
@@ -354,9 +357,26 @@ func (druid *Druid) applyScarletEnclaveGuardian6PBonus() {
...
+	})
+
+	core.MakeProcTriggerAura(&druid.Unit, core.ProcTrigger{
+		Name:     label,
+		Callback: core.CallbackOnSpellHitDealt,
+		Outcome:  core.OutcomeCrit,
+		Handler: func(sim *core.Simulation, spell *core.Spell, result *core.SpellResult) {
+			savageFlurryAura.Activate(sim)
+		},
+	})
```
kayla-glick, line 379: https://wago.tools/db2/SpellAuraOptions?build=1.15.7.60013&filter%5BSpellID%5D=1226126&page=1
https://wowdev.wiki/Spell.dbc/procFlags

Looking at the proc flag on this, it appears to work on both direct and periodic crits without restrictions. So we probably want to make this a custom aura with both callbacks

kayla-glick (approved):

PR #1432, opened 2025-04-02
Fix the 6p from the latest ptr update, add a property on the rogue proto to be able to update the dynamic value of t4 2p tank

fynnen:
image

### DIFF
--- a/sim/rogue/item_sets_phase_8.go
+++ b/sim/rogue/item_sets_phase_8.go
@@ -137,13 +137,14 @@ func (rogue *Rogue) applyScarletEnclaveTank2PBonus() {
 	if rogue.HasAura(label) {
 		return
 	}
+	rogue.rollingWithThePunchesDamageMultiplier += 0.01
 
 	rogue.RegisterAura(core.Aura{
 		Label: label,
 		OnInit: func(aura *core.Aura, sim *core.Simulation) {
 			rogue.RollingWithThePunchesProcAura.ApplyOnStacksChange(func(aura *core.Aura, sim *core.Simulation, oldStacks int32, newStacks int32) {
-				rogue.PseudoStats.DamageDealtMultiplier /= 1 + 0.01*float64(oldStacks)
-				rogue.PseudoStats.DamageDealtMultiplier *= 1 + 0.01*float64(newStacks)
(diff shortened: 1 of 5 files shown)


sim/rogue/item_sets_phase_8.go
```diff
@@ -218,9 +219,9 @@ func (rogue *Rogue) applyScarletEnclaveTank6PBonus() {
 		Label: label,
 		OnInit: func(aura *core.Aura, sim *core.Simulation) {
 			rogue.RollingWithThePunchesProcAura.ApplyOnStacksChange(func(aura *core.Aura, sim *core.Simulation, oldStacks int32, newStacks int32) {
-				if newStacks == 10 {
+				if newStacks == 5 {
 					energyAura.Activate(sim)
-				} else if newStacks < 10 && oldStacks == 10 {
+				} else if newStacks < 5 && oldStacks == 5 {
```
kayla-glick, line 224: I think we can just make this `else if oldStacks == 5` since if it's not `== 5` then it must be `> 5`

kayla-glick, line 224: I think we can just make this `else if oldStacks == 5` since if it's not `== 5` then it must be `< 5`

```suggestion
				} else if oldStacks == 5 {
```

kayla-glick (approved):

fynnen:
Rolling with the punches health changed to 5% per stack (was 6%)
T4 6p only goes up to 5 stacks (was 10).
T4 6p causes you to gain 2% more health per stack of Rolling with the punches (total of 35% fully stacked)
T4 6p causes you to gain an additional 1% damage per stack from the 2pc (max 10% damage from the effect so unchanged at max stacks but not takes 5 stacks instead of 10)
T4 6p effect now activates on Dodges OR Parries (was just Dodges)
T4 6p effect when at max stacks reduced to 10 energy restored on dodge/parry (was 15 energy)
Main Gauche's duration changed to 5 seconds (was 10)
Main Gauche's sinister strike buff chnaged to 20 second duration (was 30)
Main Gauche's Sinister Strike buff causes hits of Sinister Strike to apply a stack of Rolling with the Punches

PR #1433, opened 2025-04-03
standardize WS batching between classes

Adamrch:


Adamrch:
WS/WF ICD intentionally left at 1.5s, while the AP buff was changed to 1.5 + spell batch delay

kayla-glick (approved):

PR #1437, opened 2025-04-04
Add: shadow_priest Phase 7 + 8

Aftermath:
This PR adds BIS builds from @Shiromar1122 for Phase 7 + 8, and incorporates them into shadow_priest presets and sim.
Also sets Phase 8 as default selections.

Please let me know if I have misinterpreted some configuration here.

### DIFF
--- a/sim/priest/shadow/shadow_priest_test.go
+++ b/sim/priest/shadow/shadow_priest_test.go
@@ -128,6 +128,8 @@ func TestShadow(t *testing.T) {
 			EPReferenceStat: proto.Stat_StatSpellPower,
 			StatsToWeigh:    Stats,
 		},
+		core.GetTestBuildFromJSON(proto.Class_ClassPriest, 7, 60, "../../../ui/shadow_priest/builds", "phase_7", ItemFilters, proto.Stat_StatSpellPower, Stats),
+		core.GetTestBuildFromJSON(proto.Class_ClassPriest, 8, 60, "../../../ui/shadow_priest/builds", "phase_8", ItemFilters, proto.Stat_StatSpellPower, Stats),
 	}))
 }
 
--- a/ui/shadow_priest/presets.ts
+++ b/ui/shadow_priest/presets.ts
@@ -1,26 +1,6 @@
 import { Phase } from '../core/constants/other.js';
 import * as PresetUtils from '../core/preset_utils.js';
 import { Spec } from '../core/proto/common';
-import {
-	Conjured,
-	Consumes,
-	Debuffs,
-	EnchantedSigil,
-	Flask,
-	Food,
-	IndividualBuffs,
-	ManaRegenElixir,
-	Potions,
-	Profession,
-	RaidBuffs,
-	SaygesFortune,
-	ShadowPowerBuff,
-	SpellPowerBuff,
-	TristateEffect,
-	WeaponImbue,
-	ZanzaBuff,
-} from '../core/proto/common.js';
-import { ShadowPriest_Options as Options } from '../core/proto/priest.js';
 import { SavedTalents } from '../core/proto/ui.js';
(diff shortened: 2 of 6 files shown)


ui/shadow_priest/presets.ts
```diff
@@ -60,9 +71,11 @@ export const GearPresets = {
 	[Phase.Phase4]: [GearPhase4],
 	[Phase.Phase5]: [GearPhase5Draconic, GearPhase5CoreForged],
 	[Phase.Phase6]: [GearPhase6],
+	[Phase.Phase7]: [PresetBuildPhase7.gear!],
+	[Phase.Phase8]: [PresetBuildPhase8.gear!],
 };
 
-export const DefaultGear = GearPresets[Phase.Phase6][0];
+export const DefaultGear = GearPresets[Phase.Phase8][0];
```
kayla-glick, line 78: You should be able to remove the `DefaultGear`, `DefaultTalents`, and `DefaultAPL` constants now and instead, in `sim.ts` pull directly from `DefaultBuild`! See the `ui/mage/sim.ts` for an example. I actually also fully removed buffs, debuffs, and consumes from the `presets.ts` in favor of pulling directly from the `DefaultBuild`

kayla-glick (changes requested):
Thanks for contributing! Would you mind also adding test cases in `sim/priest/shadow/shadow_priest_test.go`? They should be just 1 line of code each. The `dps_mage_test.go` has examples of it!

ui/shadow_priest/presets.ts
Aftermath: Okay, I'll try to cleanup and align to the mage config.

ui/shadow_priest/sim.ts
Aftermath, line 1: @kayla-glick the only thing I am unsure about here is `raidSimPresets`. I updated items there to use DefaultBuild, but the mage `sim.ts` has an empty `raidSimPresets`. Should I be removing its contents here as well?

sim/priest/shadow/shadow_priest_test.go
```diff
@@ -128,6 +128,8 @@ func TestShadow(t *testing.T) {
 			EPReferenceStat: proto.Stat_StatSpellPower,
 			StatsToWeigh:    Stats,
 		},
+		core.GetTestBuildFromJSON(proto.Class_ClassPriest, 7, 60, "../../../ui/shadow_priest/builds", "phase_7", ItemFilters, proto.Stat_StatSpellPower, Stats),
+		core.GetTestBuildFromJSON(proto.Class_ClassPriest, 8, 60, "../../../ui/shadow_priest/builds", "phase_8", ItemFilters, proto.Stat_StatSpellPower, Stats),
```
Aftermath, line 132: I presume this is all I need in the way of tests?

ui/shadow_priest/sim.ts
kayla-glick: @Aftermath you can pretty much ignore those. They're used for a full raid sim that we tested out in Wrath that just wasn't worth how much work it took to maintain it

sim/priest/shadow/shadow_priest_test.go
kayla-glick: Yep that should work! It pulls all of the settings from the JSON

Aftermath:
I think this is good, I'm guessing github still wants changes because I didn't force push the previous commits and instead added new ones?

Anything else you would like me to change?

Thanks.

kayla-glick (approved):

kayla-glick:
Oh woops yeah I merged master just to be safe and forgot to come back to it

kayla-glick:
You _might_ have some test conflicts with some tuning changes that just went out so I'm double checking it again

PR #1438, opened 2025-04-04
BUGFIX - Sebacious Poison Applied In Settings > Debuffs Counts Towards P8 Rogue DPS Tier 2pc Bonus

Puddingstomp:
WIP but I am attempting to add a new callback to Auras to differentiate whether or not they "come from" the Settings UI in the sim. This should fix the interaction where if Sebacious Poison is turned on in Settings > Debuffs, it counts towards the Rogue P8 DPS Tier 2pc bonus when it shouldn't.

This has some ramifications when it does get fixed, as the SSL APL that Comfy currently has set up is relying on this bug to make the "pk swap macro" work.


kayla-glick:
Got an alternate fix in place for this. I basically wrapped the poisons that could be applied via settings in a wrapper aura and adjusted the counts in the wrapper instead

PR #1455, opened 2025-04-27
Hunter - Adding movement to hunter pets

ncberman:
Hunter pet AI now moves into melee range when outside of it
Hunter pets now start the fight at the same position as the hunter

Added a guard to prevent MoveTo() being executed when unit is already moving (caused move aura stacks to go negative when MoveTo was called while moving)

Added HunterPetIsActive to APL options
Added HunterPetFocus to APL options
Added HunterPetFocusPercent to APL options


sim/hunter/apl_values.go
```diff
@@ -0,0 +1,98 @@
...
+	return &APLValueHunterPetIsActive{
+		hunter: hunter,
+	}
+}
+
+func (value *APLValueHunterPetIsActive) Type() proto.APLValueType {
+	return proto.APLValueType_ValueTypeBool
+}
+func (value *APLValueHunterPetIsActive) GetBool(sim *core.Simulation) bool {
+	return value.hunter.pet != nil
```
kayla-glick, line 38: We could go a step further here and say `value.hunter.pet != nil && value.hunter.pet.IsActive()` since pets can be deactivated

PR #1461, opened 2025-05-03
Ravagane fixes

JJwang11:
[Logs](https://sod.warcraftlogs.com/reports/GZ6DKxjVQcHB7nad?fight=last&type=damage-done&source=1) confirm:
- 9 second buff duration, AoE every 1.5 seconds (First tick is 1.5 seconds after buff is obtained)
- Can use abilities during it
- Can proc Weapon Enchants (03:12.533 Bladestorm hit, 03:12.533 Righteous Inquisition Refresh, no other hits)
- Can proc itself on last Bladestorm tick after 8 sec ICD (03:12.533 Bladestorm hit, 03:12.532 Bladestorm Buff is removed and Applied)
- If Bladestorm buff is refreshed while active, the next tick is still 1.5 seconds after refresh (03:03.525 Bladestorm is refreshed, 03:02.437 -> 03:05.084 for next Bladestorm tick)
- Can store + proc extra attacks (00:34.322 Bladestorm fades, 0.34.953 Double Melee removing 2 Wild Strike stacks), maybe there is retry timer? Need to also set max stored attacks for Hunter at 2.

There is currently a bug with this implementation where sometimes AutoAttack does not turn back on after Bladestorm fades. Not sure where it is.

### DIFF
--- a/sim/common/sod/item_effects/phase_8.go
+++ b/sim/common/sod/item_effects/phase_8.go
@@ -754,7 +754,7 @@ func init() {
 			SpellSchool: core.SpellSchoolPhysical,
 			DefenseType: core.DefenseTypeMelee,
 			ProcMask:    core.ProcMaskMeleeMHSpecial,
-			Flags:       core.SpellFlagNoOnCastComplete | core.SpellFlagPassiveSpell | core.SpellFlagSuppressEquipProcs,
+			Flags:       core.SpellFlagNoOnCastComplete | core.SpellFlagPassiveSpell,
 
 			DamageMultiplier: 1,
 			ThreatMultiplier: 1,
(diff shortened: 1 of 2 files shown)


sim/common/sod/item_effects/phase_8.go
```diff
@@ -789,13 +788,32 @@ func init() {
...
+
+				whirlwindSpell.AOEDot().Cancel(sim)
 				character.AutoAttacks.CancelAutoSwing(sim)
-				channelSpell.AOEDot().Apply(sim)
+				whirlwindSpell.AOEDot().Apply(sim)
 			},
 			OnExpire: func(aura *core.Aura, sim *core.Simulation) {
 				character.AutoAttacks.EnableAutoSwing(sim)
-				channelSpell.AOEDot().Cancel(sim)
+				whirlwindSpell.AOEDot().Cancel(sim)
```
kayla-glick, line 816: I know that you mentioned it was sometimes not starting to attack again. Does swapping the order of these fix that?

kayla-glick (approved):

PR #1467, opened 2025-05-06
ravagane fix

Adamrch:
Fixes proc mask on whirlwind to suppress Equip effects
Fixes autos going out too soon after whirlwind ends
Allows WS attack to go out immediate after proccing Bladestorm on same swing.
Fixes all cancelaura Issues breaking Ravagane proc when used in APL.
Also fixes auto attacks just being allowed to be re-enabled during whirlwind.

### DIFF
--- a/sim/common/sod/item_effects/phase_8.go
+++ b/sim/common/sod/item_effects/phase_8.go
@@ -748,13 +748,17 @@ func init() {
 	// https://www.wowhead.com/classic/item=240919/ravagane
 	// Chance on hit: You attack all nearby enemies for 9 sec causing weapon damage plus an additional 200 every 1.5 sec.
 	// Confirmed PPM 0.8
-	itemhelpers.CreateWeaponProcAura(Ravagane, "Ravagane", 0.8, func(character *core.Character) *core.Aura {
+
+	core.NewItemEffect(Ravagane, func(agent core.Agent) {
+		character := agent.GetCharacter()
+		tickPeriod := time.Millisecond * 1500
+
 		tickSpell := character.RegisterSpell(core.SpellConfig{
 			ActionID:    core.ActionID{SpellID: 1231546},
 			SpellSchool: core.SpellSchoolPhysical,
 			DefenseType: core.DefenseTypeMelee,
 			ProcMask:    core.ProcMaskMeleeMHSpecial,
-			Flags:       core.SpellFlagNoOnCastComplete | core.SpellFlagPassiveSpell,
+			Flags:       core.SpellFlagNoOnCastComplete | core.SpellFlagPassiveSpell | core.SpellFlagSuppressEquipProcs,
 
 			DamageMultiplier: 1,
 			ThreatMultiplier: 1,
@@ -769,7 +773,6 @@ func init() {
 		})
 
 		whirlwindSpell := character.RegisterSpell(core.SpellConfig{
-			ActionID:    core.ActionID{SpellID: 1231547},
 			SpellSchool: core.SpellSchoolPhysical,
 			ProcMask:    core.ProcMaskMeleeMHSpecial,
 			Dot: core.DotConfig{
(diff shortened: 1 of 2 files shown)


Adamrch:
Cancel Aura issue example: Autos dont restart if cancelAura is used in APL
image

Adamrch:
Here's an example caused by auto attack not being disabled just toggled off previously.
image

kayla-glick:
@Adamrch can you explain why we need the `#AllowAutoSwing` method and the `allowed` state versus what exists now?

sim/core/attack.go
```diff
@@ -293,6 +293,7 @@ type WeaponAttack struct {
 	curSwingSpeed    float64
 	curSwingDuration time.Duration
 	enabled          bool
+	allowed          bool
```
kayla-glick, line 296: @Adamrch can you explain why we need the `#AllowAutoSwing` method and the `allowed` state versus what exists now?

sim/common/sod/item_effects/phase_8.go
```diff
@@ -785,36 +788,60 @@ func init() {
...
+					NumTicks:        6,
+					Priority:        core.ActionPriorityDOT,
+					TickImmediately: false,
+					OnAction: func(sim *core.Simulation) {
+						if aura.IsActive() && aura.TimeActive(sim) >= tickPeriod { // Prevent previous proc from casting on Refresh
+							aura.RemoveStack(sim)
+							whirlwindSpell.AOEDot().TickOnce(sim)
+						}
+					},
+				})
```
kayla-glick, line 821: Why do we need to use a separate periodic action here versus just using the AOEDot's aura?

sim/common/sod/item_effects/phase_8.go
```diff
@@ -785,36 +788,60 @@ func init() {
...
-				aura.Icd.Use(sim)
+				// Delay Disabling of Auto to allow Extra attacks to go out
+				core.StartDelayedAction(sim, core.DelayedActionOptions{
+					DoAt: sim.CurrentTime + time.Millisecond * 50, // Exact amount of time unknown
+					OnAction: func(s *core.Simulation) {
+						if aura.IsActive() {
+							character.AutoAttacks.AllowAutoSwing(sim, false)
+						}
+					},
+				})
```
kayla-glick, line 808: This is a problem if a swing timer happens to fall within this window right?

sim/common/sod/item_effects/phase_8.go
```diff
@@ -785,36 +788,60 @@ func init() {
...
+			SpellFlagsExclude: core.SpellFlagSuppressWeaponProcs, // Trigger is Chance on Hit so suppress if triggering spell does not proc Chance on Hit
+			PPM:               0.8,
+			ICD:               8 * time.Second,
+			Handler: func(sim *core.Simulation, _ *core.Spell, result *core.SpellResult) {
+				whirlwindAura.Activate(sim)
+				whirlwindAura.SetStacks(sim, 6)
+			},
+		})
+
+		character.ItemSwap.RegisterProc(Ravagane, triggerAura)
```
kayla-glick, line 844: Good call on converting this 👍 totally missed that the first time around. It feel like we would benefit from making a helper for CoH aura procs

sim/common/sod/item_effects/phase_8.go
```diff
@@ -785,36 +788,60 @@ func init() {
...
-			Label:    "Ravagane Bladestorm",
-			Duration: time.Second * 9,
-			Icd: &core.Cooldown{
-				Timer:    character.NewTimer(),
-				Duration: time.Second * 8,
-			},
+		whirlwindAura := character.RegisterAura(core.Aura{
+			Label:    "Ravagane Bladestorm Disable Autos",
+			ActionID:    core.ActionID{SpellID: 1231547},
+			Duration: time.Second * 9 + core.SpellBatchWindow, // Allows 6th and final tick
```
kayla-glick, line 794: I'm actually thinking that we should not expose this aura to the APL and should instead expose the AOEDoT's aura. Then have that aura control the disabling/re-enabling of autos and cancelling the aura in the APL both removes the DoT so that it won't hit, plus it will then go ahead and re-enable autos. I don't really see a need for this aura to exist when we can and should be using that one?

sim/common/sod/item_effects/phase_8.go
Adamrch: Possibly, we don't know exactly at what point between 0 and 1.5 the auto attack gets disabled, but it's not disabled immediately after the first hit. Suggestions?

sim/core/attack.go
Adamrch: What exists now is the player controlled startattack / stopattack. Allowed is a just a flag that causes the swing timer to reset and no auto to go out as seen in game.  (the delay at the end of the whirlwind for the first auto is consistant with the swingtimer filling and reseting during the whirlwind)

sim/common/sod/item_effects/phase_8.go
Adamrch: I can try to implement again using the AOEDot Aura but dot damage was not going away at the right time when bladestorm aura was canceled in my branch.  (could you help take a look and replace periodic action with the AoEDot apply() ?)  I dont mind changing that if it works but it seemed to have issues in my branch.

sim/common/sod/item_effects/phase_8.go
Adamrch: I couldn't get much of anything to work right without using this aura. A ton of issue popped up for me when try to avoid it.  If you can get the original set of auras to work better and fix the issues I pointed out I would gladly go for it.  (I can't really explain why I need this but I struggled to fix the issues without it)

sim/common/sod/item_effects/phase_8.go
Adamrch: thanks

Adamrch:
image

Adamrch:
Seed:  3116164887
https://wowsims.github.io/sod/retribution_paladin/#eJztVUFIFFEYnv/N7O6bJwzPqXT25cE2imFBmd1tRV1zJwm1hWihAo9WKBpRopc6ZZBkQiUdcrMSwUMmaBJCEBJlBgodKikstUwvUQlmlw5hvZldtzpsdrAI2m9g+P/v/9/3v/nfm/dIDkYUGVAIJqqGGIJm1A2tCNoRDCIoy8VAuYs06LUJE2YBomgQyutgDKASVIHdp8QZPXLgRE0jFbWSXMl7A0gGvTKhaA9C+kSMEYkOTCmcejWkaI+268vDEe40Tynap5A+u8S408WTZ0L63EeNOx2Wo+iTMStylTtLIf3LokYwvTSt6AOnLLozrj62aA24xp3Xit7PI5jODSl69zxwq2VK0S+e3sjLn+PlJXqev12046WinSklAifeX1f827BMFd1pNwAqpXaEjrtawdEOvA1SL8A9gKfgpDxcCZMgvQVYBChelsgsEtL4o1Bj6RanwDvU4/AIqz3e9Db9C0hv05Qoy6KoEPgpyo9WiEIMoBrqoAEi7qBhBA1fXtAI5AX9/oARCPr8hhHwBaNQJR0UDo+KmkgoEftA8MjERRx5vvxgk5rDGMnyrsegyiPgJBKeWYgwBxEJBIoFlTDMnPHjXpWZywrgkZPqFraZuL3ZWFTlO9Yg+mJaYZhwy9JkCUZlTCPrvJlYUsU2SYgnGDyBT0Hdw3aTUrWEyLfsqrMLEZLt3cBT44oWYekQ5GtKNT9PBpH7bI7fcqqH5aZKZCs5+9lesotVELlTsqi7ozmkWC1MNcz+RD6hZI2VTyxoImIbCLbszS5Y6RLmV2rSftPnTtpW6XK2kxSqBbamY0WTJ31v288hrvVDR7mFh3tAdzSMwWcE44n/A5tPElaROYmgEwR+szIJixQ887JsR5przcx4Tr7p7rxsYTxcFGcehzfV2nge9l8Q8cMPiuqqqDla01h/SAt7/7GDTj9rronO1v5VdCp+Gafs69rMg8Nnvxtum0X1xxZbxvc9C+9IRMwqSK9IEukV+U38byvyDa8k/Mg=

PR #1482, opened 2025-06-29
Add Regicide aura for APL usage

catszeid:
Add an aura for Regicide to denote when the alternate conditions for Execute are met. This makes it so the APL can more easily recognize when the effect is triggered and mirror the game implementation.

### DIFF
--- a/sim/warrior/items.go
+++ b/sim/warrior/items.go
@@ -278,7 +278,12 @@ func (warrior *Warrior) ApplyRegicideWarriorEffect(itemID int32, aura *core.Aura
 		})
 	})
 	// Regicide aura for APL use on trigger from stacks
-	regicideAura := warrior.NewTemporaryStatsAura("Regicide", core.ActionID{SpellID: 1231436}, stats.Stats{}, 10*time.Second)
+	regicideAura := warrior.RegisterAura(
+		core.Aura{
+			ActionID: core.ActionID{SpellID: 1231436},
+			Label:    "Regicide",
+			Duration: time.Second * 10,
+		})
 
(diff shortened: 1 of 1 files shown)


sim/warrior/items.go
```diff
@@ -315,13 +318,18 @@ func (warrior *Warrior) ApplyRegicideWarriorEffect(itemID int32, aura *core.Aura
 		Handler: func(sim *core.Simulation, spell *core.Spell, result *core.SpellResult) {
 			debuff := debuffAuras.Get(result.Target)
 			debuff.Activate(sim)
+			// Activate aura for APL usage when effect is triggered
+			// Implementation of cast condition is below in 'ApplyOnInit'
+			if debuff.GetStacks() == 20 {
+				regicideAura.Activate(sim)
+			}
```
kayla-glick, line 325: I imagine that we'd want to do this after the `AddStack` below? this would only activate the aura after proccing again after reaching 20 on the previous proc

sim/warrior/items.go
```diff
@@ -277,6 +277,8 @@ func (warrior *Warrior) ApplyRegicideWarriorEffect(itemID int32, aura *core.Aura
 			Duration:  time.Second * 15,
 		})
 	})
+	// Regicide aura for APL use on trigger from stacks
+	regicideAura := warrior.NewTemporaryStatsAura("Regicide", core.ActionID{SpellID: 1231436}, stats.Stats{}, 10*time.Second)
```
kayla-glick, line 281: Since this isn't an aura that affects stats we can just do a simple `RegisterAura` instead of the empty stats like thid

